PR #43, opened 2022-03-05
Add method to integrate 1D models

lpsinger:


m4opt/models/math.py
```diff
@@ -0,0 +1,247 @@
...
+def integrate(model: Model, *,
+              quick_and_dirty_npts: Optional[int] = None,
+              **kwargs) -> Union[Quantity, float]:
+    """Integrate a 1D model using adaptive trapezoidal quadrature.
+
+    Parameters
+    ----------
+    model : :class:`astropy.modeling.Model`
+        The model to integrate.
+    quick_and_dirty_npts : int, optional
```
eteq, line 172: maybe
```suggestion
    non_adaptive_npts : int, optional
```
or something like that? "quick and dirty" is in some ways accurate, but it took me some time to understand what you really meant here algorithmically

m4opt/models/math.py
```diff
@@ -0,0 +1,247 @@
...
+        if hasattr(result, 'to_value'):
+            result = result.to_value()
+        return result
+
+    if quick_and_dirty_npts is not None:
+        x = np.linspace(a, b, quick_and_dirty_npts)
+        yint = np.trapz(func(x), x)
+    else:
+        yint, _ = quad_vec(func, a, b, points=points,
+                           quadrature='trapezoid', **kwargs)
```
eteq, line 245: maybe
```suggestion
        kwargs.setdefault('quadrature', 'trapezoid')
        yint, _ = quad_vec(func, a, b, points=points, **kwargs)
```
just so that the user has the option to re-set this if they really want to?

m4opt/models/math.py
```diff
@@ -0,0 +1,247 @@
...
+            bbox = model.bounding_box
+        except NotImplementedError:
+            unit = _get_1d_units_for_dict(model.input_units)
+            return portion.open(-np.inf * unit, np.inf * unit)
+        else:
+            (lo, hi), = bbox
+            return portion.open(lo, hi)
+
+
+def integrate(model: Model, *,
```
eteq, line 163: this only works on  models that have a single `x` dimension as written, it seems to me.  Maybe this should be:
```suggestion
def integrate1d(model: Model, *,
```
or similar? I know this is the majority of model cases, but it's still important to clarify that it doesn't work for e.g. the 2D image models

m4opt/models/math.py
```diff
@@ -0,0 +1,247 @@
...
+    """
+    x_unit = _get_1d_units_for_dict(model.input_units)
+
+    intervals = _get_intervals(model)
+    a, *points, b = np.unique(Quantity(list(chain.from_iterable(
+        (i.lower, i.upper) for i in intervals)), x_unit).value.ravel())
+
+    # FIXME: Cannot rely on model.return_units because it is not set for all
+    # models (e.g. Lorentz1D).
+    y_unit = getattr(model(a * x_unit), 'unit', None) or 1
```
eteq, line 232: This is not intended behavior, right?  It's probably a bug in Astropy?  If so we should probably make sure it gets reported.

eteq (approved):
Looks good overall.  Several minor suggestions, but nothing's a deal-breaker so I'm approving contingent on @lpsinger looking at those and either accepting or ignoring them.

m4opt/models/math.py
lpsinger: Actually I prefer "quick and dirty" because it forces the user to accept that there are no guarantees of accuracy.

If we keep it called `quick_and_dirty_npts`, can we improve the documentation to make it more clear what it does differently?

m4opt/models/math.py
lpsinger: Yes, I think it's a bug in Astropy. With https://github.com/astropy/astropy/pull/12926 I was trying to fix the easiest cases.

m4opt/models/math.py
lpsinger: Done.

Issue #47, opened 2022-04-29
Risks with lack of Internet access?


cbard:
As discussed in PR #45, there are some `astropy` commands that require a internet connection to work, namely `SkyCoord.from_name` and `EarthLocation.of_site`. (see: https://docs.astropy.org/en/stable/coordinates/remote_methods.html) . 

Since we anticipate that `m4opt` may be used in places without internet (or restricted access environments), we need to document any risks with using these methods in the `m4opt` workflow, and be aware of such places in `m4opt`.

We may also consider writing a script to enable "offline mode", a la https://docs.astropy.org/en/stable/utils/iers.html#working-offline .

cbard:
Addressed by #48 and #49.

PR #48, opened 2022-05-03
Added risk tracking page to docs

cbard:
This is my attempt to encapsulate the risk tracking matrix in the html documentation, as required by NPR 7150.2.


cbard:
This will also address #47; risks for internet access and for not having MILP solver licenses are documented.

PR #50, opened 2022-05-10
Airmass and Extinction Objects

cbard:
This is a reopening of #45. The original PR was accidentally closed due to weird consequences after making m4opt public.


m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,338 @@
...
+        else:
+            self.model = airmass_models[model_name.lower()]
+
+    def at(self, target_coord, obs_time):
+        """
+        returns airmass at target sky location at obs_time
+        """
+        if not isinstance(target_coord, SkyCoord):
+            raise TypeError("argument target_coord must be of type \
+                            astropy.coordinates.sky_coordinate.SkyCoord")
```
eteq, line 147: I don't think this should check explicitly for `SkyCoord`, because this machinery should also work for e.g. `AltAz` or `ICRS` objects.  Instead I suggest duck-typing, something like 
```suggestion
        if not hasattr(target_coord, 'transform_to'):
            raise TypeError("argument target_coord must be an astropy coordinates object")
```

eteq (changes requested):
Several inline comments, but one bigger one: there are no uinit tests!  This is complex enough that that is critical.

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,338 @@
...
+    table = QTable.read(download_file(path_to_extin+tablename, cache=True),
+                        format='ascii', names=('wavelength', 'extinction'))
+
+    x = (table['wavelength']*u.Angstrom).to(
+        BaseExtinction.input_units['x'], equivalencies=u.spectral())
+    y = table['extinction']
+    return np.flipud(x), np.flipud(y)
+
+
+def simple_airmass(z):
```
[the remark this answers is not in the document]
cbard: Resolved by making input argument an `astropy` coordinate object and accessing `coord.secz`.

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,338 @@
...
+    Simple airmass calculation assuming plane-parallel atmosphere.
+    Not recommended for zenith angles (z) > 75 degrees.
+
+    Input:
+    z : zenith angle (`astropy.units.Quantity`)
+    """
+    return 1./np.cos(z)
+
+
+def KastenYoung_airmass(z):
```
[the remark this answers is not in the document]
cbard: The issue is that not all coordinate objects have `zen` or `secz`, or even `coord.altaz.zen`. For example `m33 = SkyCoord.from_name('m33')` is a coordinate object (has `transform_to`), but does not have `m33.secz`.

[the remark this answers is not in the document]
lpsinger: Oh, you need to transform to altaz.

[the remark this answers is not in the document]
cbard: The `Airmass.at()` method takes care of this by transforming the input coordinate to the appropriate `AltAz` object and calling `self.model(altaz.zen)`. 

I could probably change the models to require an `AltAz` coordinate as input, but it seems simpler to take in the zenith angle (even at the cost of duplicating effort for the simple model).

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,338 @@
...
+    >>> import astropy.units as u
+    >>> from astropy.coordinates import SkyCoord, EarthLocation
+    >>> from astropy.time import Time
+
+    >>> place_airm = Airmass(EarthLocation(lat=41.3*u.deg, lon=-74*u.deg,\
+                             height=390*u.m))
+
+    or choose a pre-defined location, e.g. Kitt Peak:
+
+    >>> kpno_airm = Airmass.kpno()
```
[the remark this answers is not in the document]
cbard: Ok, that's fine. I will change to steer users towards the `Earth_Location` functionality.

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,339 @@
...
+    """
+
+    def __init__(self, earth_location, airmass_model=None):
+        if not isinstance(earth_location, EarthLocation):
+            raise TypeError("Input earth_location must be of type" +
+                            "astropy.coordinates.earth.EarthLocation")
+
+        self.earth_loc = earth_location
+        self.set_model(airmass_model)
+        return
```
lpsinger, line 131: Return is implicit at the end of a function.

```suggestion
```

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,339 @@
...
+    """Returns airmass state with unknown observatory location, which
+    will need to be set as part of the ObservingState
+    """
+    observing_state = state.get()
+    airmass = Airmass(observing_state.observatory_loc)
+    return airmass.at(observing_state.target_coord,
+                      observing_state.obstime) * u.dimensionless_unscaled
+
+
+class AtmoExtinction:
```
lpsinger, line 189: Rename to `AtmosphericExtinction`

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,339 @@
...
+    """
+    AtmoExtinction: Attentuation of spectrum due to atmospheric effects.
```
lpsinger, line 191: No need to repeat the name of the class in the first line of the docstring.

```suggestion
    Attentuation of spectrum due to atmospheric effects.
```

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,339 @@
...
+
+    Atmospheric effects are calculated via an airmass extinction table,
+    and so require the observatory location to be defined for airmass
+    calculation.
+
+    NOTE: Default extinction table is Kitt Peak.
```
lpsinger, line 197:
```suggestion
    Notes
    -----

    Default extinction table is Kitt Peak.
```

lpsinger (changes requested):

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,339 @@
...
+
+    @classmethod
+    def kpno(cls):
+        """
+        defines Airmass Object using location of Kitt Peak National Observatory
+
+        Last updated via EarthLocation.of_site("Kitt Peak") on 2022/04/27
+        """
+        return cls(EarthLocation(-1994502.60430614, -5037538.54232911,
+                                 3358104.99690298, unit='m'))
```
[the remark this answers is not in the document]
cbard: I can move this function to the unit tests and delete it here.

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,334 @@
...
+    Simple airmass calculation assuming plane-parallel atmosphere.
+    Not recommended for zenith angles (z) > 75 degrees.
+
+    Input:
+    z : zenith angle (`astropy.units.Quantity`)
+    """
+    return 1./np.cos(zen)
+
+
+def KastenYoung_airmass(zen):
```
eteq, line 50: Looking at the latest version I now see the point that if we are treating this as a primarily *internal* function, and the user interface is through the `Airmass` class, then there's really no need for using coordinate objects here.

I do however, suggest adding an `__all__` to clarify this - see my comment just after this one.

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,334 @@
...
+from astropy.coordinates import EarthLocation, AltAz
+from astropy.modeling import custom_model
+from astropy.modeling.models import Const1D, Tabular1D
+from astropy.table import QTable
+from astropy.utils.data import download_file
+
+path_to_extin = ('https://raw.githubusercontent.com/astropy/' +
+                 'specreduce-data/main/specreduce_data/' +
+                 'reference_data/extinction/')
+
```
eteq, line 15: this suggestion goes with the next one I'm about to make
```suggestion

__all__ = ('Airmass', 'AtmosphericExtinction',)

```

m4opt/models/extinction/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+extinction table. They take the sky coordinate of the target as input,
+and return an attentuation factor (dimensionless) as output.
+
+Airmass models are defined at the observatory location and return the airmass
+for a given target in the sky.
+"""
+
+from .airmass import Airmass, AtmosphericExtinction
+
+__all__ = ('Airmass', 'AtmosphericExtinction',)
```
eteq, line 15:
```suggestion
from .airmass import *
```
this goes with the previous suggestion adding an `__all__ to `airmass.py`. This defers the question of "what API is actually exposed" to the `airmass.py`, which I think is cleaner because you can read it from the file where it's implemented.

eteq (approved):
I think you addressed my concerns @cbard, although I have a minor suggestion tied to cleaning up how you specify the "public" API.  But it amounts to the same either way, so I'm approving this since I'm OK with it as long as you look at my suggestion and either accept it or actively decide you don't like it.

m4opt/models/extinction/airmass.py
cbard: @eteq: gotcha, thanks for the clarification! I have also accepted Leo's suggestion above to write a `BaseAirmass` and move the implementation of these internal functions into their respective `PlaneParallelAirmass` and `KastenYoungAirmass` classes. This should clarify that these functions are indeed internal to the class.

Having done that, however, I will put `__all__` into the `airmass.py` file, since there are several functions (other than the airmass calculations) that are internal only.

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,394 @@
...
+from astropy.coordinates import EarthLocation, AltAz
+from astropy.modeling import custom_model
+from astropy.modeling.models import Const1D, Tabular1D
+from astropy.table import QTable
+from astropy.utils.data import download_file
+
+
+path_to_extin = ('https://raw.githubusercontent.com/astropy/' +
+                 'specreduce-data/main/specreduce_data/' +
+                 'reference_data/extinction/')
```
lpsinger, line 15: Fun fact: you can concatenate string literals by placing them next to each other without any operator. Just like in C!

```suggestion
path_to_extin = ('https://raw.githubusercontent.com/astropy/'
                 'specreduce-data/main/specreduce_data/'
                 'reference_data/extinction/')
```

m4opt/models/extinction/tests/test_extinction.py
```diff
@@ -0,0 +1,78 @@
...
+    assert np.isclose(atmo(3275*u.Angstrom).value, 0.01091579211750249)
+
+
+"""
+@pytest.mark.remote_data
+def test_read_extinction():
+    freq, ext = read_extinction_table("parnal")
+    assert freq[20].value == 422242898591549.25
+    assert ext[20] == 0.072
+"""
```
lpsinger, line 78: Why is this commented out?

Also "parnal" should probably be "paranal".

lpsinger (changes requested):

m4opt/models/extinction/airmass.py
cbard: TMYK (insert rainbow here)

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,338 @@
...
+        else:
+            return cls.extinction_table(table_name)
+
+    @classmethod
+    def default_table(cls):
+        """Default Airmass Extinction Table"""
+        return cls.extinction_table("kpno")
+
+    @staticmethod
+    def extinction_table(table_name):
```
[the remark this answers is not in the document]
lpsinger: [Tabular1D](https://docs.astropy.org/en/stable/api/astropy.modeling.tabular.Tabular1D.html) only supports three interpolation methods:

> The method of interpolation to perform. Supported are “linear” and “nearest”, and “splinef2d”. “splinef2d” is only supported for 2-dimensional data. Default is “linear”.

So "linear" is the highest interpolation order that will work, and it is the default for Tabular1D.

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,394 @@
...
+
+path_to_extin = ('https://raw.githubusercontent.com/astropy/'
+                 'specreduce-data/main/specreduce_data/'
+                 'reference_data/extinction/')
+
+__all__ = ('Airmass', 'AtmosphericExtinction',
+           'PlaneParallelAirmass', 'KastenYoungAirmass')
+
+
+def read_extinction_table(placename):
```
lpsinger, line 21: Please `@cache` this.

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,394 @@
...
+    extinction tables usually have
+    x in angstroms
+    y in mag per airmass
+    """
+    try:
+        tablename = AtmosphericExtinction.available_tables[placename]
+    except KeyError:
+        raise ValueError("Invalid placename {0}. Must be one of {1}".format(
+                        placename,
+                        AtmosphericExtinction.available_tables.keys()))
```
lpsinger, line 32: Suggest using an f-string.

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,394 @@
...
+                        format='ascii', names=('wavelength', 'extinction'))
+
+    x = (table['wavelength']*u.Angstrom).to(
+        BaseExtinction.input_units['x'], equivalencies=u.spectral())
+    y = table['extinction']
+    return np.flipud(x), np.flipud(y)
+
+
+class BaseAirmass:
+    """
```
lpsinger, line 44: The docstring should consist of a one-line summary, followed by a blank lines, and then the extended description.

m4opt/models/extinction/airmass.py
```diff
@@ -0,0 +1,394 @@
...
+
+    >>> extn = AtmosphericExtinction(table_name='apo', table_method='nearest')
+    >>> with state.set_observing(target_coord=target, obstime=time, \
+                                 observatory_loc=place):
+    ...     print(extn(3200*u.Angstrom))
+    0.2727180718446889
+    """
+
+    def __new__(cls, table_name=None, table_method="linear",
+                airmass_class=Airmass):
```
lpsinger, line 335: Instead of an airmass class, let's just take an airmass instance --- which could be simply a quantity, for known airmass.

m4opt/models/extinction/airmass.py
cbard: Whoa, just learned about f-strings! Yeah, that's definitely easier.

m4opt/models/extinction/airmass.py
cbard: I will go ahead and make the default `__new__()` initializer require an airmass instance rather than being a "blank" initializer. 

However, I will keep the functions that require a `Airmass` class instance, since these are required if we will have `observatory_loc` as part of the `ScienceState`. I suggest revisiting the `ScienceState` question at a later time.

m4opt/models/extinction/airmass.py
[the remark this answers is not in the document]
cbard: I will keep this argument for now. Perhaps splines will come to Tabular1D in the future?

PR #51, opened 2022-05-10
Initial attempt at Sky brightness Object

cbard:
This is a simple Sky brightness class using data taken from Neugent & Massey (2010) for Kitt Peak. 

I had issues with the procedure for converting the data (in mag/arcsec^2) into flux (ergs/sec/cm^2/angstrom/arcsec^2), so there is likely a better way to do this.


cbard:
After contacting Phil Massey, I clarified that the data was indeed in AB magnitudes. This makes it straightforward to convert to f_nu and f_lambda.

m4opt/models/background/skybright.py
```diff
@@ -0,0 +1,116 @@
...
+
+    References
+    ----------
+    .. [1] Neugent, K. and Massey, P., 2010, "The Spectrum of the Night Sky
+           Over Kitt Peak: Changes Over Two Decades". PASP, 122, 1246-1253
+           doi:10.1086/656425
+    """
+
+    if option.lower() not in kpno_sky_tables.keys():
+        raise AttributeError("option must be one of {0}".format(
```
lpsinger, line 47: Should be a `ValueError` ("Raised when an operation or function receives an argument that has the right type but an inappropriate value")

m4opt/models/background/skybright.py
```diff
@@ -0,0 +1,116 @@
...
+    not yet include effect of lunar phase.
+
+    References
+    ----------
+    .. [1] Neugent, K. and Massey, P., 2010, "The Spectrum of the Night Sky
+           Over Kitt Peak: Changes Over Two Decades". PASP, 122, 1246-1253
+           doi:10.1086/656425
+    """
+
+    if option.lower() not in kpno_sky_tables.keys():
```
lpsinger, line 46: More Pythonic ("easier to ask for forgiveness than permission"):

```python
key = option.lower()
try:
    filename = keno_sky_table[key]
except KeyError:
    raise ...
```

lpsinger (changes requested):
Very nice! Suggesting two minor changes.

m4opt/models/background/__init__.py
```diff
@@ -1,7 +1,8 @@
 from .core import Background
 from .zodiacal import ZodiacalBackground
+from .skybright import SkyBrightness
 
-__all__ = ('ZodiacalBackground',)
+__all__ = ('ZodiacalBackground', 'SkyBrightness',)
```
lpsinger, line 5: Do we want to call this `SkyBackground`, for consistency with the other classes in this module?

lpsinger (changes requested):

m4opt/models/background/__init__.py
cbard: Yes, we should do that.

Issue #55, opened 2022-05-27
Add example code for constructing bandpasses


lpsinger:
Document how to construct bandpass models. We need to support the following use cases:

1. User provides tabular data
2. User selects pre-defined bandpass by name (note, [tynt](https://tynt.readthedocs.io) is coming soon)
3. User builds arbitrary model from other 1D models (e.g. Gaussian, compound model, etc.)

PR #56, opened 2022-06-01
Bandpass models

cbard:
Addresses #55. Adds a Bandpass Object class to construct from:

1. user-provided arrays
2. 1D astropy model classes and their constructor arguments
3. user-provided tynt filter name

Some things to be aware of:

- The `tynt` package does not exist, so `tynt` is currently an optional dependency. Tox tests are provided, but are not enabled.
- I used code from `synphot.spectrum.SourceSpectrum` to handle the 1D Model argument handling (and put the appropriate BSD-3 license in the license folder).


m4opt/models/bandpass/core.py
```diff
@@ -0,0 +1,15 @@
+from astropy.modeling import Model
+from astropy import units as u
+
+
+class Background(Model):
```
lpsinger, line 5: Probably meant to call this `Bandpass`?

m4opt/models/bandpass/bandpass.py
```diff
@@ -0,0 +1,202 @@
...
+from astropy import units as u
+from astropy.units.physical import get_physical_type
+
+from .core import Background
+
+# TODO: remove check below when tynt is released
+# and add tynt to required dependencies
+try:
+    from tynt import FilterGenerator
+    has_tynt = True
```
lpsinger, line 12: Remove `has_tynt` flag; no need for it because instead can just check if `tynt_filters is None`.

m4opt/models/bandpass/bandpass.py
```diff
@@ -0,0 +1,202 @@
...
+                                                input_units_equivalencies)
+            return result
+
+        else:
+            print("optional dependency 'tynt' is not installed. " +
+                  "Will return 'None'")
+            return None
+
+    @classmethod
+    def from_model(cls, modelclass, **kwargs):
```
lpsinger, line 114: Initializing a model from a model? I don't get it...

lpsinger (changes requested):

m4opt/models/bandpass/bandpass.py
cbard: This was my attempt to satisfy the criteria "User builds arbitrary model from other 1D models (e.g. Gaussian, compound model, etc.)" from #55. This imitates `synphot.SourceSpectrum` by taking in a `Model` class, and checking whether it has the appropriate input/output units. It also prevents using an inappropriate model for the Bandpass.

cbard: That's fine. But in that case, we don't need to write any code to handle that situation? Any arbitrary `Model1D` can be compounded with any other model. All we would have to do is document that the input must have spectral units, and the output must be dimensionless.

cbard: `from_model()` here does work as intended. 

If you're talking about taking a generic model and multiplying it with, e.g., `ZodiacalBackground`, then yes, it works (as long as the units are right)

lpsinger: Then I think that we can just drop the `from_model()` method --- and the LICENSE file.

m4opt/models/bandpass/core.py
lpsinger: Sure.

m4opt/models/bandpass/bandpass.py
```diff
@@ -0,0 +1,115 @@
...
+
+from .core import BaseBandpass
+
+# TODO: remove check below when tynt is released
+# and add tynt to required dependencies
+try:
+    from tynt import FilterGenerator
+    tynt_filters = FilterGenerator().available_filters()
+except ImportError:
+    tynt_filters = None
```
eteq, line 13: It's not clear `tynt` will *ever* get a true release, as it's slated to be added to synphot - see https://github.com/spacetelescope/synphot_refactor/pull/257 .  So as I see it for now there's two options:
1.  add the *specific git commit of tynt* to the `setup.cfg` file and treat that as a hard dependency for now. You can see how to do that here: https://python-packaging.readthedocs.io/en/latest/dependencies.html#packages-not-on-pypi
2. Have a clearer optional dependency message - see my immediately following suggestion:

m4opt/models/bandpass/bandpass.py
```diff
@@ -0,0 +1,115 @@
...
+            filter = FilterGenerator().reconstruct(filter_name)
+            result = Tabular1D(filter.wavelength,
+                               filter.transmittance*u.dimensionless_unscaled)
+            result.input_units_equivalencies = (BaseBandpass.
+                                                input_units_equivalencies)
+            return result
+
+        else:
+            print("optional dependency 'tynt' is not installed. " +
+                  "Will return 'None'")
```
eteq, line 114: You should not use un-supressable prints for something like this - in general `print` is not a reliable tool for use in a library.  Moreover, you *definitely* do not want to  return None where the user expected to get something real.  This is exactly when you should raise an exception.  So I'd say do something like:
```suggestion
            raise ImportError("optional dependency 'tynt' is not installed. To install it, do `pip install git+https://github.com/bmorris3/tynt.git@master`" )
```

Also, FYI, the `+` is unnecessary here as you originally wrote it- Python automatically concatenates multiline strings in this context.

m4opt/models/bandpass/bandpass.py
```diff
@@ -0,0 +1,115 @@
...
+
+        result = Tabular1D(points, lookup_table*u.dimensionless_unscaled)
+        result.input_units_equivalencies = (BaseBandpass.
+                                            input_units_equivalencies)
+        return result
+
+    @classmethod
+    def from_name(cls, filter_name):
+        """
+        Initializes bandpass from tynt package.
```
eteq, line 97:
```suggestion
        Returns a bandpass for a filter specified by the name ``filter_name``, using the `tynt <https://github.com/bmorris3/tynt>`_ package's parametrizations .
```

eteq (changes requested):
A few inline comments. 

But I am still missing something: why aren't we just using the synphot bandpass objects: https://synphot.readthedocs.io/en/latest/synphot/bandpass.html ?  Those seem to be *exactly duplicative of this (but with a lot more features). They even include the tynt models as of last week (although that's not in the release yet).

cbard:
@eteq I think that's a fair question. I actually did have code copied from `synphot.SourceSpectrum` in an earlier iteration of this `m4opt` class. I was assuming that @lpsinger didn't want `synphot` as a hard dependency, but I may be mistaken.

@lpsinger : Any thoughts on this?

bmorris3:
Hi folks! I'm looking to do an update to `tynt` ASAP. I need it as a dependency for another project, so I'm going to remove astropy-helpers, and modernize the repo in general. I see the workflow in this PR depends on tynt, so I wanted to give you a heads up.

lpsinger:
Thanks @bmorris3! Will be handy.

bmorris3:
First tynt release is out on [PyPI](https://pypi.org/project/tynt/0.1.0/) and [GitHub](https://github.com/bmorris3/tynt/releases/tag/v0.1.0).

PR #57, opened 2022-06-09
Tick NPR 7150 secure coding standard box 🙄

lpsinger:
Instead of LGTM, we enabled GitHub's CodeQL scanning.

Yay, security posturing.


eteq:
Sort of 6 one half dozen of the other, but just to clarify: is it that LGTM isn't sufficient for NPR 7150 and CodeQL is?

lpsinger:
No. They both do the same thing. The difference is that there is a button in the GitHub project settings to turn on CodeQL, so I did that.

PR #64, opened 2023-09-30
Add more documentation for zodiacal light background

lpsinger:
- Warn that it is only valid near Earth.
- Add plots.
Screenshot 2023-09-30 at 16 48 15


lpsinger:
@bwgref, do you know why I wouldn't see the two light bands above and below the ecliptic, like we did in the background plots from the site visit?

mcoughlin (approved):
Good warning message.

bwgref:
> @bwgref, do you know why I wouldn't see the two light bands above and below the ecliptic, like we did in the background plots from the site visit?

@lpsinger The UVEX ETC also includes the models for the galactic diffuse emission, which can dominate over the Zodi when you're close-ish to the Plane.

PR #71, opened 2023-10-02
Rework SkyBackground docs

lpsinger:
- Remove docstring from undocumented read_kpno_sky_data. It's not visible in the Sphinx docs, and most of its contents was repeated.
- Add a pretty plot.
Screenshot 2023-10-01 at 20 12 28


mcoughlin (approved):
Yay pretty.

PR #73, opened 2023-10-30
Add CPLEX installation details for HPC resources

bfhealy:
This PR adds details to the documentation for academic users installing CPLEX onto an HPC cluster like SDSC Expanse.

### DIFF
--- a/docs/solvers.rst
+++ b/docs/solvers.rst
@@ -40,11 +40,11 @@ __ https://www.ibm.com/academic
 
 4. Follow the download and installation instructions.
 
-   If running on a cluster like SDSC Expanse, copy the `linux_x86_64`_ installer
-   file (e.g. `cplex_studio2211.linux_x86_64.bin`_) to the cluster. Then,
+   If installing on a remote system, be sure to copy the appropriate installer
+   file (e.g. `cplex_studio2211.linux_x86_64.bin`) to that system. Then,
    update the permissions of the file to make it executable and run it 
-   with `./cplex_studio2211.linux_x86_64.bin`_. The installation will then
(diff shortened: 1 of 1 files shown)


docs/solvers.rst
```diff
@@ -40,6 +40,12 @@ __ https://www.ibm.com/academic
 
 4. Follow the download and installation instructions.
 
+   If running on a cluster like SDSC Expanse, copy the `linux_x86_64`_ installer
+   file (e.g. `cplex_studio2211.linux_x86_64.bin`_) to the cluster. Then,
+   update the permissions of the file to make it executable and run it 
+   with `./cplex_studio2211.linux_x86_64.bin`_. The installation will then
+   run in the command line.
```
lpsinger, line 47: This isn't specific SDSC Expanse, or even specific to HPC clusters, is it? This applies to any Linux system.

bfhealy: This is indeed applicable to any Linux system - Michael and I were thinking it would be helpful to highlight the additional steps needed to install CPLEX on a cluster given that the CPLEX download will likely happen locally. 

Maybe I should rephrase the above the be more generic, e.g. "For installation on a remote Linux system..."

lpsinger: It's not specific to remote installation either, is it?

bfhealy: I think the part about copying the installer file is specific to a remote setup. While it's a small extra step, a version of this PR's added documentation may be useful for a user who's otherwise used to running `module load` or using pip/conda to install software on a cluster.

lpsinger (approved):

PR #172, opened 2024-09-19
Rename MILP decision variables to use notation from design doc

lpsinger:
This makes the constraint expressions a lot shorter. Are they more readable this way? Or do the short variable names make the code
more confusing?


lpsinger:
Although this is cute, I think it mostly detracts from readability. What do you think, @mcoughlin?

mcoughlin:
@lpsinger Maybe we put something in the design documentation where we give the variables?

lpsinger:
Yeah, or in the comments for this function.

PR #275, opened 2025-02-21
Implement AirmassConstraint for FoR limitation

weizmannk:
This PR adds a Field of Regard (FoR) constraint using AirmassConstraint to ensure observations are limited based on airmass values.

### DIFF
--- a/m4opt/constraints/_airmass.py
+++ b/m4opt/constraints/_airmass.py
@@ -1,5 +1,3 @@
-from typing import Optional
-
 import numpy as np
 from astropy import units as u
 
@@ -19,7 +17,7 @@ class AirmassConstraint(AltitudeConstraint):
     ----------
     max_airmass : float
         Maximum airmass of the target (corresponds to minimum altitude).
-    min_airmass : float, optional
+    min_airmass : float,
         Minimum airmass of the target (corresponds to maximum altitude).
         Default is `1` (the zenith).
 
@@ -45,7 +43,7 @@ class AirmassConstraint(AltitudeConstraint):
     def __init__(
         self,
         max_airmass: float,
-        min_airmass: Optional[float] = 1,
+        min_airmass: float= 1.0,
     ):
         min_alt = np.arcsin(1 / max_airmass) * u.rad
         max_alt = np.arcsin(1 / min_airmass) * u.rad
--- a/m4opt/constraints/_airmass.py
+++ b/m4opt/constraints/_airmass.py
@@ -43,7 +43,7 @@ class AirmassConstraint(AltitudeConstraint):
     def __init__(
         self,
         max_airmass: float,
-        min_airmass: float= 1.0,
+        min_airmass: float = 1.0,
     ):
         min_alt = np.arcsin(1 / max_airmass) * u.rad
         max_alt = np.arcsin(1 / min_airmass) * u.rad
--- a/m4opt/constraints/_airmass.py
+++ b/m4opt/constraints/_airmass.py
@@ -61,9 +61,3 @@ class AirmassConstraint(AltitudeConstraint):
 
         super().__init__(min=min_alt, max=max_alt)
 
-    @override
-    def __call__(self, observer_location, target_coord, obstime):
-        """
-        Compute the airmass constraint by leveraging altitude constraints.
-        """
-        return super().__call__(observer_location, target_coord, obstime)
--- a/m4opt/constraints/_airmass.py
+++ b/m4opt/constraints/_airmass.py
@@ -57,11 +57,7 @@ class AirmassConstraint(AltitudeConstraint):
             if min_airmass is None
             else Angle(np.degrees(np.arcsin(1 / min_airmass)) * u.deg)
         )
-        max_alt = (
-            Angle(np.nan * u.deg)
-            if max_airmass is None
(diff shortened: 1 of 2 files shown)


lpsinger (changes requested):
- [ ] Remove `min_best_rescale` method.
- [ ] Subclass `AltitudeConstraint`; convert the minimum and maximum airmass values to altitudes.

weizmannk (commented):
LGTM

weizmannk:
@lpsinger In order to address this requested change, I need to merge the altitude branch (Add AltitudeConstraint to limit FoR #277) into this one. This will be easy for  CI/CD pipeline. 

I mean, for the subclass AltitudeConstraint, it will make more sense to merge both branches

```
from .altitude import **AltitudeConstraint**
class AirmassConstraint(**AltitudeConstraint**):
```
If you agree, I will merge them together and then push the changes I made.
Thanks

lpsinger:
> If you agree, I will merge them together and then push the changes I made.

I think it would be better to finish and merge the AltitudeConstraint PR first.

lpsinger:
> > If you agree, I will merge them together and then push the changes I made.
> 
> I think it would be better to finish and merge the AltitudeConstraint PR first.

Merged! Please rebase, then update this PR.

m4opt/constraints/_airmass.py
```diff
@@ -0,0 +1,73 @@
...
+
+    The airmass is approximated as the secant of the zenith angle, which allows
+    for an equivalent formulation in terms of altitude constraints. This class
+    extends `AltitudeConstraint` and internally converts the given airmass
+    limits into corresponding altitude limits.
+
+    Parameters
+    ----------
+    min_airmass : float or `None`, optional
+        Minimum airmass of the target. Default is `1` (the zenith).
```
lpsinger, line 23: What is the use case for a minimum airmass constraint?

m4opt/constraints/_airmass.py
```diff
@@ -0,0 +1,73 @@
...
+    Notes
+    -----
+    - The conversion from airmass to altitude follows the relation:
+        `altitude = arcsin(1 / airmass)`, which assumes a standard atmosphere.
+    """
+
+    def __init__(
+        self,
+        min_airmass: Optional[float] = 1,
+        max_airmass: Optional[float] = None,
```
lpsinger, line 53: `max_airmass` probably should not be optional.

m4opt/constraints/_airmass.py
```diff
@@ -0,0 +1,73 @@
...
+        min_alt = (
+            Angle(90 * u.deg)
+            if min_airmass is None
+            else Angle(np.degrees(np.arcsin(1 / min_airmass)) * u.deg)
+        )
+        max_alt = (
+            Angle(np.nan * u.deg)
+            if max_airmass is None
+            else Angle(np.degrees(np.arcsin(1 / max_airmass)) * u.deg)
+        )
```
lpsinger, line 64:
```suggestion
        max_alt = np.arcsin(1 / max_airmass) * u.rad
```

m4opt/constraints/_airmass.py
```diff
@@ -0,0 +1,73 @@
...
+
+        super().__init__(min=min_alt, max=max_alt)
+
+    @override
+    def __call__(self, observer_location, target_coord, obstime):
+        """
+        Compute the airmass constraint by leveraging altitude constraints.
+        """
+        return super().__call__(observer_location, target_coord, obstime)
```
lpsinger, line 73:
```suggestion
```

lpsinger (changes requested):

m4opt/constraints/_airmass.py
```diff
@@ -0,0 +1,51 @@
...
+    >>> location = EarthLocation.of_site("Rubin Observatory")
+    >>> constraint = AirmassConstraint(max_airmass=3, min_airmass=1)
+    >>> constraint(location, target, time)
+    np.True_
+    """
+
+    def __init__(
+        self,
+        max_airmass: float,
+        min_airmass: float = 1.0,
```
lpsinger, line 46: Why do we need to support a minimum airmass? I don't see the use case.

lpsinger (changes requested):

m4opt/constraints/_airmass.py
mcoughlin: In the past, we have wanted to take calibration standards at a variety of air masses, so one could use this to ensure some high airmass observations.

lpsinger: I see. Does it still make sense that minimum airmass is optional?

lpsinger: Sorry, what I mean to ask is: does it still make sense that maximum airmass is required?

mcoughlin: I think a sensible default is ok.

lpsinger: Can we add:

```python
min_airmass, max_airmass = sorted([min_airmass, max_airmass])
```

so that it doesn't matter the order in which these arguments are passed? That would allow us to support both of the following:

```python
AirmassConstraint(2.5)
AirmassConstraint(3, 4)
```

lpsinger (changes requested):

lpsinger (approved):

PR #280, opened 2025-02-27
Add Earth-fixed observer class

lpsinger:


m4opt/orbit/__init__.py
mcoughlin, line 1: Perhaps it is time to rename orbit to e.g. coordinates?

mcoughlin (approved):
Just a question on whether now is a good time to rename orbit.

m4opt/orbit/__init__.py
lpsinger: Please comment on https://github.com/m4opt/m4opt/issues/281.

PR #282, opened 2025-03-01
Rename Orbit to ObserverLocation

### LINKED ISSUE #281: Rename m4opt.orbit because we now support observing from locations that are not on orbits
We now support earth-fixed observers, so `m4opt.orbit.Orbit` is no longer a suitable module name. What should we call it? [In Astroplan, this is called an Observer.](https://astroplan.readthedocs.io/en/stable/api/astroplan.Observer.html) However, the Observer class has a lot of additional methods.

lpsinger:
Fixes #281.


mcoughlin (approved):
LGTM

PR #285, opened 2025-03-04
Add LSST mission details

weizmannk:


lpsinger (commented):
These pages are not currently included in the docs build. It would be better to create an LSST mission object and add this information to its docstring.

weizmannk:
@lpsinger 
OK, but I recall you mentioned that you would be handling the development of the API for ground-based telescopes, because  they are a bit more complex, maybe i missed something. I am not sure if the LSST mission can be created as simply as ULTRASAT or UVEX.? If yes I could start fill out the parameters.  Furthermore, I still have not had access to all the information, we are missing details such as read_noise, dark_noise, gain, SNR, the sensitivity area, and possibly the Sun–Moon separation angle we need to avoid.

I will ask in the email thread who can provide a link to obtain that information.

lpsinger:
See https://smtn-002.lsst.io.

lpsinger:
Closing this in favor of #292.

PR #286, opened 2025-03-06
Refactor airmass and altitude constraints

lpsinger:
- Both classes now support passing the minimum and maximum in any order.
- The airmass constraint now supports both float and quantity input.

### DIFF
--- a/m4opt/constraints/_airmass.py
+++ b/m4opt/constraints/_airmass.py
@@ -4,6 +4,25 @@ from astropy import units as u
 from ._altitude import AltitudeConstraint
 
 
+def airmass_to_altitude(
+    airmass: float | u.Quantity[u.physical.dimensionless],
+) -> u.Quantity[u.physical.angle]:
+    """Convert airmass to altitude using the cosecant formula.
+
+    Examples
+    --------
+    You can pass either floating-point values or dimensionless values with units.
+
+    >>> from astropy import units as u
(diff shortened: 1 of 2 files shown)


m4opt/constraints/_airmass.py
```diff
@@ -4,6 +4,29 @@
...
+    >>> from astropy import units as u
+    >>> from m4opt.constraints._airmass import airmass_to_altitude
+    >>> airmass_to_altitude(1)
+    <Quantity 1.57079633 rad>
+    >>> airmass_to_altitude(1 * u.dimensionless_unscaled)
+    <Quantity 1.57079633 rad>
+    >>> airmass_to_altitude(2.5)
+    <Quantity 0.41151685 rad>
+    >>> airmass_to_altitude(2.5 * u.dimensionless_unscaled)
+    <Quantity 0.41151685 rad>
```
weizmannk, line 25: Do we need to repeat the example for both the minimum and maximum values? If not, I suggest using either 1 or 2.5, but not both.
```suggestion
```

weizmannk (changes requested):

m4opt/constraints/_airmass.py
lpsinger: I deleted the airmass 1 example.

weizmannk (approved):
LGTM

Issue #274, opened 2025-02-21
Integration on Ground-Based Telescope: ZTF and LSST


weizmannk:
We are gathering the essential elements regarding the constraints as well as the background noise.  

## Priority Constraints: 
- [x] **AltitudeConstraint**  
- [x] **AirmassConstraint**  
- [x] **AtNightConstraint**  

> **Related Branch:**  
> [ground-telescope-config](https://github.com/m4opt/m4opt/compare/main...weizmannk:m4opt:ground-telescope-config)

##  Additional Constraints (Potentially Useful): 
- [ ] **MoonIlluminationConstraint**  
- [ ] **LocalTimeConstraint**  
- [ ] **PrimaryEclipseConstraint**  
- [ ] **SecondaryEclipseConstraint**  
- [ ] **Constraint**  
- [ ] **TimeConstraint**  

An old issue already exists for the [Reference scenario for ZTT](https://github.com/m4opt/m4opt/issues/30) . To keep the missions separate, we could open a new issue specifically for LSST.  
This would allow us to track mission-specific requirements for both projects more effectively.

lpsinger:
@weizmannk, would you please open PRs to add each of the constraints one at a time? It will be a lot easier for me to review that way.

weizmannk:
Okay great, I will do it.

weizmannk:
Hi @lpsinger ,
I have two questions about the LSST mission.
1- Similar to the ZTF case, observations using the ToO process follow a specific cadence defined by the strategy.
Should we add this as a constraint, or do we intend to implement it during the scheduling process?
Here is an example of the strategy required for an NS-NS event. 

The section P**referred strategy**,  page 8 , https://arxiv.org/pdf/2111.01945 

2. Do we want to use the LocalTimeConstraint, which specifies a specific time for the observation?
For example, this could be something like '23:00 to 05:00.

Here are the missions parameters : [LSST Mission Parameters](https://github.com/weizmannk/m4opt/blob/LSST-mission-params/m4opt/missions/_lsst.py). 
I still need to double-check the slew acceleration, as this value seems odd to me compared to the old paper.
`
max_angular_acceleration=5.25 * u.deg / u.s**2`

lpsinger:
[...]

PR #292, opened 2025-03-10
Add new mission: LSST

### LINKED ISSUE #302: New mission configuration: LSST

weizmannk:
Fixes #302.


m4opt/missions/_lsst.py
```diff
@@ -0,0 +1,96 @@
...
+from ..synphot.background import GalacticBackground, ZodiacalBackground
+from ._core import Mission
+
+lsst = Mission(
+    name="lsst",
+    fov=CircleSkyRegion(center=SkyCoord(0 * u.deg, 0 * u.deg), radius=1.75 * u.deg),
+    constraints=[
+        AirmassConstraint(max=2.5, min=1),
+        AltitudeConstraint(min=20 * u.deg, max=85 * u.deg),
+        AtNightConstraint.twilight_civil(),
```
lpsinger, line 26:
```suggestion
```

Redundant because astronomical twilight is more constraining.

m4opt/missions/_lsst.py
```diff
@@ -0,0 +1,96 @@
...
+        AtNightConstraint.twilight_astronomical(),
+        MoonSeparationConstraint(30 * u.deg),
+    ],
+    detector=Detector(
+        npix=6,  # https://github.com/lsst/rubin_sim/blob/2bf176a6d98ff4c84c352912c5e0721e330fc217/rubin_sim/skybrightness/sky_model.py#L144C19-L144C26
+        plate_scale=(0.2 * u.arcsec) ** 2,
+        # Circular aperture with a diameter of 6.423 m
+        area=np.pi * np.square(0.5 * 6.423 * 100 * u.cm),
```
lpsinger, line 34:
```suggestion
        area=np.pi * np.square(0.5 * 6.423 * u.m),
```

m4opt/missions/_lsst.py
```diff
@@ -0,0 +1,96 @@
...
+                amplitude=0.69,
+                mean=8750 * u.angstrom,
+                stddev=750 * u.angstrom,
+            ),
+            "ps1__y": SpectralElement(
+                Gaussian1D,
+                amplitude=0.35,
+                mean=10000 * u.angstrom,
+                stddev=1000 * u.angstrom,
+            ),
```
lpsinger, line 71: Accurate LSST filter bandpasses are definitely public knowledge, and there is no need to use Gaussian approximations. I don't know what the official reference is for LSST, but here are some general-purpose sources that will at least supply the PanSTARRS and SDSS filter bandpasses:

- https://astroquery.readthedocs.io/en/latest/svo_fps/svo_fps.html
- https://stsynphot.readthedocs.io/en/latest/stsynphot/appendixb_nonhst.html#stsynphot-appendixb-nonhst
- https://github.com/bmorris3/tynt (although this last one may be unmaintained now; @bmorris3, what should we be using instead?)

m4opt/missions/_lsst.py
```diff
@@ -0,0 +1,96 @@
...
+    ),
+    # The LSST (Vera C. Rubin Observatory) is a ground-based telescope
+    observer_location=EarthFixedObserverLocation.of_site("LSST"),
+    # Sky grid optimized for LSST's large field of view.
+    skygrid=skygrid.geodesic(9.6 * u.deg**2, class_="III", base="icosahedron"),
+    # Slew model tailored for LSST (Vera C. Rubin Observatory), a ground-based telescope in Chile.
+    slew=Slew(
+        max_angular_velocity=6.3 * u.deg / u.s,
+        max_angular_acceleration=5.25 * u.deg / u.s**2,
+        settling_time=15 * u.s,
```
lpsinger, line 86: Where do these numbers come from?

lpsinger (changes requested):

m4opt/missions/_lsst.py
weizmannk: The speed was calculated using information from [this LSST document](https://docushare.lsstcorp.org/docushare/dsweb/Get/Publication-76/).

 The typical slew rate (TSR) is based on:
- An azimuth rate of 7 deg s⁻¹ over a range of 360°
- An elevation rate of 3.5 deg s⁻¹ over a range of 90°

Because of the larger range in azimuth, a weighted average with a 4:1 ratio is used:

\[
\text{TSR}  = \frac{(4 \times \text{Azimuth Rate}) + \text{Elevation Rate}}{5}
            = \frac{(4 \times 7) + 3.5}{5}
            = 6.3\,\text{deg.s}^{-1}
\]
This value of  6.3\,\text{deg.s}^{-1} have been used in the gwemopt too. 

The acceleration was taken from [this TMA document](https://project.lsst.org/reviews/tma/sites/default/files/docs/Document-14411.pdf). However, the maximum acceleration is not clear to me  as noted in [this issue #274](https://github.com/m4opt/m4opt/issues/274).

lpsinger: Please capture this in the comments.

m4opt/missions/_lsst.py
```diff
@@ -0,0 +1,96 @@
...
+)
+from ..dynamics import Slew
+from ..observer import EarthFixedObserverLocation
+from ..synphot import Detector
+from ..synphot.background import GalacticBackground, ZodiacalBackground
+from ._core import Mission
+
+lsst = Mission(
+    name="lsst",
+    fov=CircleSkyRegion(center=SkyCoord(0 * u.deg, 0 * u.deg), radius=1.75 * u.deg),
```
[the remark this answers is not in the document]
lpsinger: The `fov` property can be a [`Regions`](https://astropy-regions.readthedocs.io/en/stable/api/regions.Regions.html#regions.Regions) object that consists of a collection or regions, so you could do what I did for ZTF (#293) and model each of the chips or each of the rafts as a convex polygon.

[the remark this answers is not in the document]
weizmannk: @lpsinger 
This is when I used the Regions object to create a collection of LSST FOVs (a total of three rectangles).
[Plus Sign](https://github.com/weizmannk/EarthOrbitPlan/blob/main/script_from_leo_paper.png)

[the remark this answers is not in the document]
lpsinger: Does the LSST camera have gaps between its chips or rafts? If so, it would be good to model them. It's unlikely that we will evaluate HEALPix footprints finely enough to resolve the gaps, but even so it would be nice to be able to see the gaps in plots of the vector regions themselves.

Even if there are no gaps between the rafts, would you please model each raft as a separate polygon? It will look nicer in plots.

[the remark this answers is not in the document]
weizmannk: @lpsinger That's one look more fine 

LSST FOV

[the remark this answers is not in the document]
lpsinger: That's better. I can't find any documentation online about the width of the LSST chip and raft gaps. @mcoughlin, who would be a good person to ask for those specifications?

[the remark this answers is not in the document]
lpsinger: @weizmannk, we got pointed to [`lsst.obs.lsst.LsstCam`](https://pipelines.lsst.io/py-api/lsst.obs.lsst.LsstCam.html`) for the authoritative model of the LSST camera geometry.

m4opt/missions/_lsst.py
lpsinger: You can now use [`bandpass_from_svo`](https://m4opt.readthedocs.io/en/latest/api/m4opt.synphot.bandpass_from_svo.html#m4opt.synphot.bandpass_from_svo) to retrieve [all LSST transmission curves from SVO](http://svo2.cab.inta-csic.es/svo/theory/fps/index.php?mode=browse&gname=LSST).

rhiannonlynne: You can do that - I don't know what version they pulled. 
You can also look at https://github.com/lsst/throughputs

m4opt/missions/_lsst.py
rhiannonlynne: How much detail can you capture here? 
We have an entire observatory model in rubin_scheduler that will tell you what the slewtime is between any two points, including if we have to change the filter. 
One issue is that the TMA is faster than the Dome, for long slews, but there is a dome crawl feature that lets the dome move to some extent during exposures (assuming that the next exposure will be nearby). 
If you want the full model, see here - https://github.com/lsst/rubin_scheduler/blob/main/rubin_scheduler/scheduler/model_observatory/kinem_model.py

But if you just need a simple set of numbers to plug in that will probably be good enough, please use the current defaults of 
```        altitude_minpos=20.0,
        altitude_maxpos=86.5,
        azimuth_minpos=-250.0,
        azimuth_maxpos=250.0,
        altitude_maxspeed=3.5,
        altitude_accel=3.5,
        altitude_jerk=14.0,
        azimuth_maxspeed=7.0,
        azimuth_accel=7.0,
        azimuth_jerk=28.0,
        settle_time=3.0,
```
(velocity is deg/second, acceleration in deg/second^2, and jerk is deg/second^^3).

m4opt/missions/_lsst.py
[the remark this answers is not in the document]
rhiannonlynne: We do have a set of vertices in a simple file for LsstCam, but the pointer to obs_lsst above is more definitive.

m4opt/missions/_lsst.py
weizmannk: @rhiannonlynne  Thank you for these information.s  
However, there are some discrepancies between the values depending on the paper and the repositories.  
Perhaps it would be better if we had an **official paper** that we could reference.  

I would like to confirm the actual field of view of LSST. Is it **3.5 degrees** or **9.6 degrees**?  
[FOV](https://github.com/lsst/rubin_scheduler/blob/main/rubin_scheduler/scheduler/model_observatory/kinem_model.py#L174)  

@lpsinger, I assume it would be more complex to take into account the **dome configuration** and the **slew**?  
[dome setup](https://github.com/lsst/rubin_scheduler/blob/main/rubin_scheduler/scheduler/model_observatory/kinem_model.py#L228-L229)

lpsinger:
Please rebase.

weizmannk:
I already rebased, but something didn't follow up correctly.  Maybe I can create a new branch, maybe  more clean.

lpsinger:
> I already rebased, but something didn't follow up correctly. Maybe I can create a new branch, maybe more clean.

I rebased it for you and force-pushed to your branch.

m4opt/missions/_lsst.py
lpsinger: > How much detail can you capture here?

@rhiannonlynne, we currently have [a toy slew model which carries out an eigenaxis maneuver with a maximum angular rate, maximum angular acceleration, and fixed settling time](https://m4opt.readthedocs.io/en/latest/api/m4opt.dynamics.EigenAxisSlew.html#m4opt.dynamics.EigenAxisSlew). It doesn't currently support different dynamics about different axes, and doesn't have a maximum jerk term. This model is definitely too simple to accurately model real spacecraft and real ground-based telescope mounts.

It's an open question how much detail we need to add to make a model that is "good enough." For spacecraft, the dynamics may be hard to model and/or proprietary, so in some cases we may need to be able to describe a dynamical envelope that encloses the true agility of the spacecraft with some margin. For ground-based telescopes, especially large international projects like LSST, a lot more detail is usually known.

In theory, we could create a subclass of our [generic slew model](https://m4opt.readthedocs.io/en/latest/api/m4opt.dynamics.Slew.html#m4opt.dynamics.Slew) that calls out to a third-party package such as the one you referred us to. However, for now, I'd like to avoid adding large dependencies.

My main request to @weizmannk was to make sure that wherever these parameters come from, the code comments clearly document the primary source that we drew them from.

> However, there are some discrepancies between the values depending on the paper and the repositories. Perhaps it would be better if we had an official paper that we could reference.

@weizmannk, I would be perfectly happy with treating the Rubin scheduler as a primary source. Just please update the numbers and make sure the comments explain where they came from.

> I would like to confirm the actual field of view of LSST. Is it 3.5 degrees or 9.6 degrees?

It's 3.5 degrees on a side, with an area of 9.6 square degrees.

> @lpsinger, I assume it would be more complex to take into account the dome configuration and the slew?

Interesting question. For GW ToOs where we frequently have multiple widely separated islands of probability, it is more important to accurately model long slews than short ones, so that the scheduler employs long slews only sparingly. So I would suggest using the dome dynamics here.

I wonder if this is a common situation for ground-based observatories, that the telescope mount is more agile than the dome. Can we do quick survey of some of our favorite telescopes (P48, P200, Gemini, Keck, LSST) to determine if this is the case?

m4opt/missions/_lsst/data/lsst_science_detectors.csv
lpsinger, line 1: Where is this file from, and under what license did we receive it? You may need to convey it with a copy of the LICENSE file, like I have here: https://github.com/m4opt/m4opt/pull/293/files#diff-4edb7884c5a6febf7ae48fec99d91f6bad769b590d0640befc5a37f97fbd6c63

m4opt/missions/_lsst/lsst_camera.py
```diff
@@ -0,0 +1,82 @@
...
+
+from astropy import units as u
+from astropy.coordinates import SkyCoord
+from astropy.table import Table
+from regions import PolygonSkyRegion, RectangleSkyRegion, Regions
+
+from . import data
+
+
+class LSSTfieldOfView:
```
lpsinger, line 20: Not sure why this is a class.

m4opt/missions/_lsst/lsst_camera.py
lpsinger, line 1: If you want to split the LSST sources across multiple files, then please arrange them like this:

```
m4opt/missions/_lsst/__init__.py  # mission definition
m4opt/missions/_lsst/camera.py  # FOV geometry
m4opt/missions/_lsst/data/...  # vendored copies of external data
m4opt/missions/_lsst/data/LICENSE.txt  # README file explaining the provenance of any vendored data
m4opt/missions/_lsst/data/README.txt  # license file for any vendored data
```

lpsinger (changes requested):

m4opt/missions/_lsst/data/lsst_science_detectors.csv
weizmannk: I explained it in issue [#302](https://github.com/m4opt/m4opt/issues/302).
I am not yet sure if this is the final camera geometry, which is why I have not provided a copy of the license yet.

m4opt/missions/_lsst/lsst_camera.py
weizmannk: I can also  do it like this : http://github.com/weizmannk/EarthOrbitPlan/blob/main/lsst_fov/lsst_geometry_camera.py

m4opt/missions/_lsst/lsst_camera.py
weizmannk: It's done. To pass the CI test, I need to delete `_lsst.py` which is no longer useful, but to avoid losing the comments, I can either create an archive for them or configure pytest to ignore the file. Would it be better to archive the comments or tell pytest to ignore `_lsst.py`?

lpsinger: Why would you lose the comments? All you needed to do was rename `m4opt/missions/_lsst.py` to `m4opt/missions/_lsst/__init__.py`.

m4opt/missions/_lsst/data/lsst_science_detectors.csv
lpsinger: Can you please explain it here?

weizmannk: The `m4opt/missions/_lsst/data/lsst_science_detectors.csv` is reproduced from the following source: [`lsstCamSim.yaml`](https://github.com/lsst/obs_lsst/blob/main/policy/lsstCamSim.yaml).

1. Extracting data from `lsstCamSim.yaml` is computationally expensive.  
   To optimize processing, we extract only relevant information, focusing on science detectors  
   while excluding wavefront (WF) and guide sensors (SG). The remaining 189 science detectors are stored in an  
   `astropy.table.Table` as a CSV file (`lsst_science_detectors.csv`), which contains the LSST camera geometry.

2. This CSV file is read by `../camera.py` to construct the LSST Field of View (FOV).
(maybe @lpsinger could revise the way i write the `LSSTCameraFOV`class ,
if don't need a class I can use a simple function to deal with it). 

3. The figure (geometry camera, plus sign) should look like this one, with `nside = 1024`:  
LSST Camera FOV.

4. Unfortunately, some functions are no longer in the repository. The docstrings guide may need updates,  
such as in [`cameraGeom`](https://github.com/lsst/obs_lsst/blob/c3244459634645c954b5e7e1cb2c492cd69bc8f5/python/lsst/obs/lsst/cameraTransforms.py#L26).  

5. For more: 
To get  how I  extracted  CSV files, please see:  
[`extract_lsst_science_camera_detectors.py`](https://github.com/weizmannk/EarthOrbitPlan/blob/main/lsst_fov/extract_lsst_science_camera_detectors.py).

weizmannk: In addition it might worth to say that :  there are  two types of detector among the 189: `ITL`, with a pixel resolution of [4071, 3999], 
and `E2V`, with [4095, 4003] pixels.  And LSST features a plate scale of 0.2 arcsec/pixel with a pixel size of 0.01 mm.
I added all the "useful information" in the description.

m4opt/missions/_lsst/data/README.rst
```diff
@@ -0,0 +1,19 @@
+Reproduced Data from lsstCamSim.yaml
+====================================
+
+The contents of this directory are reproduced from the following source:
+`lsstCamSim.yaml <https://github.com/lsst/obs_lsst/blob/main/policy/lsstCamSim.yaml>`_.
+
+1. Extracting data from `lsstCamSim.yaml` is computationally expensive.
```
lpsinger, line 7: Why is it computationally expensive?

m4opt/missions/_lsst/data/README.rst
```diff
@@ -0,0 +1,19 @@
...
+   To optimize processing, we extract only relevant information, focusing on science detectors
+   while excluding wavefront and guide sensors. The remaining 189 detectors are stored in an
```
lpsinger, line 9: No, we extract only the science sensors because we only care about targets that are within the science field of view.

m4opt/missions/_lsst/data/README.rst
```diff
@@ -0,0 +1,19 @@
...
+   `astropy.table.Table` as a CSV file (`lsst_science_detectors.csv`), which contains the LSST camera geometry.
+
+2. This CSV file is read by `../camera.py` to construct the LSST Field of View (FOV).
```
lpsinger, line 12: I don't see a copy of this file in the obs_lsst repository. Is this a data product that you have generated? If so, then please include your source code.

I am optimistic that we can make that code run fast enough that we just use that, and we don't have to include the CSV file at all.

m4opt/missions/_lsst/data/README.rst
```diff
@@ -0,0 +1,19 @@
...
+
+
+3. There are two types of detectors among the 189:
+   - **ITL**: Pixel resolution of `[4071, 3999]`
+   - **E2V**: Pixel resolution of `[4095, 4003]`
```
lpsinger, line 17: Do these two types of CCDs actually have different resolutions, or simply different dimensions? If the latter, then this might not be necessary information.

m4opt/missions/_lsst/data/README.rst
```diff
@@ -0,0 +1,19 @@
...
+
+   The LSST features a plate scale of **0.2 arcsec/pixel** with a pixel size of **0.01 mm**.
```
lpsinger, line 19: The pixel size in millimeters is not necessary information.

lpsinger (changes requested):

m4opt/missions/_lsst/data/README.rst
weizmannk: it's the latter one. So I will remove it.

m4opt/missions/_lsst/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+        bandpasses={band: bandpass_from_svo(f"LSST/LSST.{band}") for band in "ugrizy"},
+        background=GalacticBackground() + ZodiacalBackground(),
+        read_noise=9,
+        dark_noise=0.2 * u.Hz,
+        gain=1,
+    ),
+    # The LSST (Vera C. Rubin Observatory) is a ground-based telescope in Chile.
+    observer_location=EarthFixedObserverLocation.of_site("LSST"),
+    # Sky grid optimized for LSST’s large field of view.
+    skygrid=skygrid.geodesic(3.5 * u.deg**2, class_="III", base="icosahedron"),
```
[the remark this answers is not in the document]
weizmannk: I do not see it yet. Maybe @rhiannonlynne  knows whether LSST has its own defined reference grid or not?

m4opt/missions/_lsst/data/README.rst
weizmannk: > The `m4opt/missions/_lsst/data/lsst_science_detectors.csv` is reproduced from the following source: [`lsstCamSim.yaml`](https://github.com/lsst/obs_lsst/blob/main/policy/lsstCamSim.yaml).
> 
> 1. Extracting data from `lsstCamSim.yaml` is computationally expensive.
>    To optimize processing, we extract only relevant information, focusing on science detectors
>    while excluding wavefront (WF) and guide sensors (SG). The remaining 189 science detectors are stored in an
>    `astropy.table.Table` as a CSV file (`lsst_science_detectors.csv`), which contains the LSST camera geometry.
> 2. This CSV file is read by `../camera.py` to construct the LSST Field of View (FOV).
>    (maybe @lpsinger could revise the way i write the `LSSTCameraFOV`class ,
>    if don't need a class I can use a simple function to deal with it).
> 3. The figure (geometry camera, plus sign) should look like this one, with `nside = 1024`:
>    LSST Camera FOV.
> 4. Unfortunately, some functions are no longer in the repository. The docstrings guide may need updates,
>    such as in [`cameraGeom`](https://github.com/lsst/obs_lsst/blob/c3244459634645c954b5e7e1cb2c492cd69bc8f5/python/lsst/obs/lsst/cameraTransforms.py#L26).
> 5. For more:
>    To get  how I  extracted  CSV files, please see:
>    [`extract_lsst_science_camera_detectors.py`](https://github.com/weizmannk/EarthOrbitPlan/blob/main/lsst_fov/extract_lsst_science_camera_detectors.py).

I have now added the script I used to produce the CSV file: [detector_extractor.py](https://github.com/m4opt/m4opt/pull/292/commits/8d422097d9f7d7eaf185d72a381a6cf7c65bc0a7).

m4opt/missions/_lsst/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+`LSST <https://rubinobservatory.org/>`_ is a ground-based optical survey telescope 
+located in Chile as part of the Vera C. Rubin Observatory. It is designed 
+to conduct a 10-year survey of the southern sky with a large field of view 
+to detect transient events, including potential gravitational wave counterparts 
+(:footcite `2019ApJ...873..111I`).
+
+The LSST camera's focal plane consists of 189 detectors arranged in 21 rafts, 
+each containing a :math:`3 \times 3` CCD, placed in a :math:`5 \times 5` grid, 
+as shown in `Figure 12  from :footcite:`2019ApJ...873..111I`. 
+We have two types of detectors: `ITL`, `E2V`.
```
lpsinger, line 65: This detail is irrelevant.

```suggestion
```

m4opt/missions/_lsst/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+LSST features a plate scale of 0.2 arcsec/pixel.
```
lpsinger, line 67: Remove; this statement adds nothing to explain the provenance of the data.

```suggestion
```

lpsinger (changes requested):

lpsinger (changes requested):
There are 29 files changes. This looks like a bad rebase.

weizmannk:
Yes, there was an issue with the ReadTheDocs process. I have reverted it and rebased it properly, I hope.

m4opt/missions/__init__.py
```diff
@@ -1,7 +1,8 @@
 """This module contains builtin settings for supported missions."""
 
 from ._core import Mission
+from ._lsst import lsst
 from ._ultrasat import ultrasat
 from ._uvex import uvex
 
-__all__ = ("Mission", "uvex", "ultrasat")
+__all__ = ("Mission", "uvex", "ultrasat", "lsst")
```
lpsinger, line 8: Sort the list:

```suggestion
__all__ = ("Mission", "lsst", "uvex", "ultrasat")
```

m4opt/missions/_lsst/camera.py
```diff
@@ -0,0 +1,115 @@
...
+import yaml
+from astropy import units as u
+from astropy.coordinates import SkyCoord
+from astropy.table import Table
+from regions import PolygonSkyRegion, RectangleSkyRegion, Regions
+
+from . import data
+
+
+class LSSTCameraFOV:
```
lpsinger, line 21: Remove the class; its methods can all be simply module-level methods.

m4opt/missions/_lsst/camera.py
```diff
@@ -0,0 +1,115 @@
...
+    def convert_rectangle_to_polygon_skyregion(rect):
+        """Convert RectangleSkyRegion to PolygonSkyRegion in RA/Dec."""
+        ra, dec = rect.center.ra.deg, rect.center.dec.deg
+        half_w = rect.width.to_value(u.deg) / 2
+        half_h = rect.height.to_value(u.deg) / 2
+        corners_ra = [ra - half_w, ra + half_w, ra + half_w, ra - half_w]
+        corners_dec = [dec - half_h, dec - half_h, dec + half_h, dec + half_h]
+        return PolygonSkyRegion(
+            vertices=SkyCoord(ra=corners_ra * u.deg, dec=corners_dec * u.deg)
+        )
```
lpsinger, line 86: I don't see why you need this method. What is wrong with leaving the regions as RectangleSkyRegion instances?

lpsinger (changes requested):

m4opt/missions/_lsst/camera.py
lpsinger, line 1: Rename to `_camera.py` to mark this as a non-public module.

lpsinger (changes requested):

m4opt/missions/_lsst/__init__.py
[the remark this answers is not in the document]
rhiannonlynne: My apologies - I missed this question. 
Is this is a reference grid for the *pointings* on the sky, as in a tesselation of fields of view, and where to point the telescope? If so, no. 
If it means, is there a reference grid that has some kind of sizing that is referred to in data products, then yes - we have "tracts" and "patches", which are used for data processing.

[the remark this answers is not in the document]
lpsinger: Ah, you're referring to LSST's dither pattern? Is there a technical document that describes that?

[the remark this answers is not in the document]
rhiannonlynne: We rotate the field tesselation on the sky, so as to get smooth dithers, rather than create offsets to dither from fixed tesselation. 
Here's an example (although not a technical document) that illustrates the spinning + tesselation
 - https://github.com/lsst-sims/scheduler_paper_apj2022/blob/main/plots/tesselation.ipynb

m4opt/missions/_lsst.py
weizmannk: @mcoughlin here is the Slew discussion : https://github.com/m4opt/m4opt/pull/292#discussion_r1999506031

mcoughlin: @weizmannk @lpsinger Is the documentation in the code now sufficient? We could start asking folks about dome performance in parallel, but this seems like a deeper dive than needed to get something working.

lpsinger:
Please fix type checking and docs build failures.

weizmannk:
@lpsinger The error comes from this line:
bandpasses  = {band: bandpass_from_svo(f"LSST/LSST.{band}") for band in "ugrizy"}

It’s not a coding issue, but rather a problem with accessing the SVO filter service, particularly for the "u" band. I have never managed to fix it reliably. Sometimes, running the code before 10 am CEST seems to work (weird, I know), maybe because the SVO or Read the Docs servers are less busy at that time. Otherwise, it often takes too long and gets rejected before the build or simulation finishes.  Unless we avoid checking it at build time, this will keep happening.

Here the full error :

`166 | sphinx.errors.ExtensionError: Handler <function process_automodsumm_generation at 0x71962530d9e0> for event 'builder-inited' threw an exception (exception: HTTPConnectionPool(host='svo2.cab.inta-csic.es', port=80): Max retries exceeded with url: /theory/fps/fps.php?ID=LSST%2FLSST.u (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x7195e9dbb010>, 'Connection to svo2.cab.inta-csic.es timed out. (connect timeout=60)'))`

PR #293, opened 2025-03-11
Add ZTF

lpsinger:


lpsinger:
I'm not aware of any existing ZTF exposure time calculator at all, are you, @mcoughlin and @weizmannk?

mcoughlin:
@lpsinger me neither. We have previously been interpolating based on the measured limiting magnitudes in NMMA.

mcoughlin:
@ebellm Do we have the specific HA constraint info somewhere from the scheduler?

ebellm:
https://github.com/ZwickyTransientFacility/ztf_sim/blob/5176ebaa9e1f8e5448593df4102d077c0e880886/ztf_sim/QueueManager.py#L1235 matches what's in ztf.rst.  Private communication from Reed Riddle.

lpsinger:
> https://github.com/ZwickyTransientFacility/ztf_sim/blob/5176ebaa9e1f8e5448593df4102d077c0e880886/ztf_sim/QueueManager.py#L1235 matches what's in ztf.rst. Private communication from Reed Riddle.

Nice, thanks! Here's a new issue for that: https://github.com/m4opt/m4opt/issues/297. Who should we assign that to?

mcoughlin (approved):
Looks good. Thanks @lpsinger

lpsinger:
I added some example ETC output to the docstring. Looks way too deep.

m4opt-missions-ztf-1

mcoughlin:
@lpsinger Where is typical seeing accounted for here?

mcoughlin:
@lpsinger I also see in the arxiv version 10 e- (https://arxiv.org/pdf/2008.04923), not that it should make substantial difference.

m4opt/missions/_ztf/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+        # Table 3 of https://ui.adsabs.harvard.edu/abs/2020PASP..132c8001D
+        plate_scale=(1.01 * u.arcsec) ** 2,
+        read_noise=8,
+        gain=6.2,
+        bandpasses={band: bandpass_from_svo(f"Palomar/ZTF.{band}") for band in "gri"},
+        background=SkyBackground.medium() + ZodiacalBackground(),
+    ),
+    observer_location=EarthFixedObserverLocation.of_site("Palomar"),
+    skygrid=_read_skygrid(),
+    # From Section 4.2:
```
ebellm, line 113: We have a more detailed model available [here](https://github.com/ZwickyTransientFacility/ztf_sim/blob/5176ebaa9e1f8e5448593df4102d077c0e880886/ztf_sim/constants.py#L27C1-L36C45) although I personally haven't validated it against the last N years of on-sky data taking.

ebellm:
Your `synphot` module doesn't seem to give you any means of specifying per-band QE/throughput--just using pi * corrector radius^2 and assuming 100% efficiency clearly is too optimistic.  One could either try to do it from first principles (CCD QE, filter and optics throughput, mirror reflectance, atmosphere...), or do it empirically by working backwards from the [metadata table](https://irsa.ipac.caltech.edu/data/ZTF/docs/ztf_metadata_latest.db).

ebellm:
There are some data files that I used for the ground-up throughput modeling [here](https://github.com/ebellm/VolumetricSurveySpeed/), although I was never totally able to get the modeled and observed limiting magnitudes to line up without some fudge factors.

m4opt/missions/_ztf/__init__.py
lpsinger: OK, it's easy enough for us to model that using separate "bang-bang" models for each Euler axis.

Does ZTF have a "dome crawl" feature like LSST, as @rhiannonlynne described in https://github.com/m4opt/m4opt/pull/292#discussion_r1997455512?

mcoughlin:
Paging through the observations page, it looks maybe a bit worse than that on average, but not enough to make up the difference here.

lpsinger:
Aha. From [Bellm et al. (2019)](https://ui.adsabs.harvard.edu/abs/2019PASP..131a8002B/abstract) I had taken the gain value of 6.2 highlighted in the screenshot below:

Screenshot 2025-03-26 at 01 47 31

I had taken this as the number of counts per photoelectron, but it is clearly stated to be the inverse, the number of photoelectrons per count.

If I set `gain=1/6.2`, then I get the plot below, which is close to the mark.

ztf

What's strange is that lots of references I have found describe gain as something that has units of e-/ADU but then the use it in a way that only makes sense if it's ADU/e-. Like for example, Astropy's own `signal_to_noise_oir_ccd` function:

https://github.com/astropy/astropy/blob/v7.0.1/astropy/stats/funcs.py#L955-L1003

@ebellm, @mcoughlin, do you agree?

ebellm:
electrons per ADU is standard, it's just that the name is not ideal.

Screenshot 2025-03-26 at 2 09 15 PM

I agree that the astropy code is confusing--the Poisson noise calculation needs to be in electrons (as readnoise is, correctly), so all the `_eps` rates need to be in ADU per second.  Then multiplying by gain in electrons/ADU gives you electrons.

lpsinger:
@ebellm, do you mean to say that the original code in which I set the gain to 6.2 is correct? Or that the version in which I set it to 1/6.2 is correct?

And where is that screen shot from?

lpsinger:
Test failures should be fixed by #354.

lpsinger:
@mcoughlin, please re-review.

mcoughlin (approved):
Great

PR #294, opened 2025-03-11
Supply default values for dark current and read noise

lpsinger:
This way, we can omit these parameters when they are negligible. For example, according to https://doi.org/10.1088/1538-3873/aaecbe:

> Dark current is negligible in maximum exposure times contemplated (300 s).


mcoughlin (approved):
Good point. LGTM

PR #296, opened 2025-03-13
Add slew base class, clear up confusion about optimal slew

lpsinger:
- Introduce a Slew base class to make room for future addition of more sophisticated slew models.
- Clear up confusion about the optimal slew trajectory. We have modeled an eigenaxis maneuver, which is _not_ the optimal maneuver in general.

### DIFF
--- a/m4opt/dynamics/_slew.py
+++ b/m4opt/dynamics/_slew.py
@@ -1,3 +1,4 @@
+from abc import ABC, abstractmethod
 from dataclasses import dataclass
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


m4opt/dynamics/_slew.py
```diff
@@ -1,25 +1,60 @@
...
+    """Model slew time for a spacecraft employing an eigenaxis maneuver.
 
-    Notes
-    -----
-    The optimal slew consists of an acceleration phase at the maximum
-    acceleration, possibly a coasting phase at the maximum angular velocity,
-    and a deceleration phase at the maximum acceleration, as shown in the
-    figure below.
+    An eigenaxis maneuver is a rotation along the path of shortest angular
+    separation, about a single axis. Assuming that the spacraft has a maximum
```
mcoughlin, line 52: spacecraft

mcoughlin (approved):
Nice. I guess other Slew models beyond EigenAxisSlew is to come.

m4opt/dynamics/_slew.py
lpsinger: Fixed.

weizmannk (approved):
LGTM

PR #303, opened 2025-03-14
Add function to get bandpass from SVO filter profile service

lpsinger:
We can use this to filter transmission curves for LSST, ZTF, etc.


weizmannk (approved):
LGTM

PR #309, opened 2025-03-18
Add CLI tool to evaluate limiting magnitude map

lpsinger:


m4opt/_cli/depth_map.py
```diff
@@ -0,0 +1,124 @@
...
+        QTable(
+            {
+                "ipix": np.concatenate(
+                    (np.arange(hpx.npix), np.concatenate(footprints))
+                ),
+                "limmag": np.concatenate((np.full(hpx.npix, -np.inf), limmag)),
+            }
+        )
+        .group_by("ipix")
+        .groups.aggregate(np.max)
```
mcoughlin, line 112: I wonder if np.max can be the default but also make np.min and np.mean as other options in case folks want to measure like "average", "best case" or "worst case"

mcoughlin (commented):
Looks good. One idea about the metric. I also asked Peter to take a quick look as he may have thoughts.

m4opt/_cli/depth_map.py
lpsinger: This is dealing with pixels that are contained in two or more overlapping fields.

tsunhopang:
Thank you very much for working on this. Would you mind providing a mock output from the code so that I can see if there is thing missing for it to be used in nmma?

tsunhopang:
Hi Leo, thank you for these example files. The format used in the test.fits.gz is good for me, and I will work on the corresponding interface in nmma.

One question, though: in the current file, there is only a single number per sky location. In case of multiple observations, should I expect one file per time, or will the file contain a time series of limit_mag?

lpsinger:
> One question, though: in the current file, there is only a single number per sky location.

That number represents the deepest observation of each pixel.

> In case of multiple observations, should I expect one file per time, or will the file contain a time series of limit_mag?

How would that work? Each observation is at a different time, so you would need a column for each observation. That would make this data product very large.

tsunhopang:
With nmma, we analyze a time series of the mag, even for non-detection limits, to impose constraints on the source parameters, etc. Of course, if it is a rare case for a non-detection event to have multiple observations across multiple days, we can also just take a single value to see what we can get from it.

lpsinger:
> With nmma, we analyze a time series of the mag, even for non-detection limits, to impose constraints on the source parameters, etc. Of course, if it is a rare case for a non-detection event to have multiple observations across multiple days, we can also just take a single value to see what we can get from it.

The limiting magnitude is different not just for each field, but for each pixel within each field. @mcoughlin, what is your vision for how you incorporate this into NMMA?

mcoughlin:
@lpsinger For NMMA, I think we should take the limiting magnitude at the point of the injection (whose location we should read from the file that specifies the object parameters) and cross-match which observations cover that particular injection location (this is harder, of course, and we need to think about how to best do that). I wonder if this step is better done as some further post-processing in m4opt (e.g. a file with observations just covering the true injection location).

PR #310, opened 2025-03-19
Wrap longitude-like angles, but not latitude-like angles

lpsinger:
- For longitude-like angles, wrap angles assuming that the caller is describing a counter-clockwise interval. For example, if max=30° and min=10°, then the constraint will return true over an interval of measure 340°.
- Only swap the order of the angles for latitude-like constraints, which are not periodic.
- Add some docstrings and change some base class names.


mcoughlin (approved):
LGTM

PR #314, opened 2025-03-26
Move DustExtinction to m4opt.synphot.extinction

lpsinger:
This will be a more reasonable location for this class because we are going to add other kinds of extinction (i.e. atmospheric extinction).


mcoughlin (approved):
LGTM

PR #316, opened 2025-04-02
Pin GitHub Actions to sha hashes

lpsinger:
See https://www.cisa.gov/news-events/alerts/2025/03/18/supply-chain-compromise-third-party-tj-actionschanged-files-cve-2025-30066-and-reviewdogaction


mcoughlin (approved):
LGTM

PR #329, opened 2025-05-12
Add radiation belt constraint

lpsinger:


weizmannk:
@lpsinger   This is what I was working on in another branch, but it was meant for the background side, so I don't need to push it again. I was expecting it to be treated as background noise rather than a constraint.  Maybe I missed something ?

lpsinger:
> @lpsinger This is what I was working on in another branch, but it was meant for the background side, so I don't need to push it again. I was expecting it to be treated as background noise rather than a constraint. Maybe I missed something ?

Yes, we still do need a background model for Cherenkov light, if the ULTRASAT team says that it is an important contributor to exposure time estimates. Please continue working on that.

This constraint class is for something completely different. Observatories in low-Earth orbit typically do not operate their science instruments while they are passing through the South Atlantic Anomaly (SAA); this constraint models that. However, ULTRASAT's orbit is high enough that it does not pass through the SAA.

PR #334, opened 2025-05-19
Cerenkov background light

weizmannk:
This model estimates the background induced by Cerenkov radiation from charged particles (electrons or protons)
  interacting with telescope optics. It uses integral flux data from the NASA AE8/AP8 radiation belt model
  to calculate the intensity of radiation emitted by typical optical materials such as silica or suprasil,
  used in the ULTRASAT telescope lenses.


weizmannk:
@lpsinger,  To ensure that the background noise is properly interpreted by m4opt, I return the final result as a spectrum: `SourceSpectrum(Empirical1D, points=wavelength, lookup_table=intensity_erg)`, similar to the one we used for the Sky Brightness background :https://github.com/m4opt/m4opt/blob/main/m4opt/synphot/background/_skybright/__init__.py#L33.

lpsinger (changes requested):
There are a large number of unrelated changes on this branch. Please rebase from main.

m4opt/synphot/background/_cerenkov/_refraction_index.py
```diff
@@ -0,0 +1,280 @@
...
+            + 0.8974794 * L**2 / (L**2 - 9.896161**2)
+        )
+        L = L * 1e4  # convert µm to Å
+        t = None
+
+    elif material == "si02_suprasil_2a":
+        # Tabulated data
+        n_data = np.array(
+            [
+                [1862, 1.5732],
```
lpsinger, line 62: Please move these large arrays to data files.

m4opt/synphot/background/_cerenkov/_refraction_index.py
```diff
@@ -0,0 +1,280 @@
...
+        Transmission per 1 cm (only for Suprasil 2A), or None.
+
+    Raises
+    ------
+    ValueError
+        For unsupported material or temperature.
+    """
+
+    material = material.lower()
+    if material in ["sio2", "silica"]:
```
lpsinger, line 46: - SiO2 is the chemical formula for silica. Do you really need to support both strings, `sio2` and `silica`?
- This function has very little in common across the three `if`/`elif` paths. Please refactor into three separate functions. If you want to provide a high-level function that provides a string-based lookup, that's fine, but please do that using a dictionary lookup.

m4opt/synphot/background/_cerenkov/_refraction_index.py
```diff
@@ -0,0 +1,280 @@
...
+
+This module computes the refractive index and transmission for optical
+materials as a function of wavelength.
+"""
+
+import astropy.units as u
+import numpy as np
+
+
+def get_refraction_index(material="si02_suprasil_2a", temperature=20):
```
lpsinger, line 14: Remove the unused `temperature` argument.

m4opt/synphot/background/_cerenkov/_electron_loss.py
```diff
@@ -0,0 +1,134 @@
...
+        plt.loglog(Ek, 1 / dEdX, linewidth=lw)
+        plt.xlabel("Electron Kinetic Energy [MeV]", fontsize=label_fs)
+        plt.ylabel(
+            r"Inverse Normalized Energy Loss $(dE/dX)^{-1}$ [(g cm$^{-2}$)/MeV]",
+            fontsize=label_fs,
+        )
+        plt.grid(True)
+        plt.axis([1e-2, 1e2, 1e-2, 1])
+        plt.tick_params(labelsize=tick_fs)
+        plt.show()
```
lpsinger, line 132: Don't include plots inside the function. This would be nice in the Example section of the docstring in a [plot directive](https://matplotlib.org/stable/api/sphinxext_plot_directive_api.html).

m4opt/synphot/background/_cerenkov/_electron_loss.py
```diff
@@ -0,0 +1,134 @@
...
+    for oxygen, silicon, and aluminum, depending on the material composition.
+    The calculations use approximate Bethe-Bloch formulas and simplified
+    radiative loss estimates.
+    """
+
+    e = 4.8e-10  # statC
+    me = 9.1e-28  # g
+    mp = 1.7e-24  # g
+    c = 3e10  # cm/s
+    hbar = 1.05e-27  # erg*s
```
lpsinger, line 53: Please use `astropy.constants` here and carry dimensions through the calculation so that we can be sure that the answer is dimensionally correct.

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,248 @@
...
+                observer_location,
+                obstime,
+                e,
+                kind="integral",
+                solar=self.solar,
+                particle=self.particle,
+            )
+            for e in energy_bins
+        ]
+        return Table([energy_bins, u.Quantity(flux_integral)], names=["energy", "flux"])
```
lpsinger, line 38: Why is this a class? It could just be a function.

lpsinger (changes requested):

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,339 @@
...
+            gamma_i = 1 + em / mass_mev
+            beta_i = np.sqrt(1 - 1.0 / gamma_i**2)
+            fCi = np.maximum(0, 1 - 1.0 / n[i] ** 2 / beta_i**2)
+            intg_i = gEE * Fm * fCi
+            cs_val = cs_intg(1.0) * intg.unit
+            Lnorm = 2 * np.pi / 137 / rho / Lam[i].to(u.cm) ** 2 * cs_val
+            Lnorm = Lnorm.to(1 / (u.MeV * u.s * u.cm**2 * u.um))
+            int_val = np.sum(intg_i * np.diff(ee)) / cs_val
+            L1mu = int_val * Lnorm
+            IC1mu[i] = L1mu.value / (2 * np.pi * n[i] ** 2)
```
lpsinger, line 133: Convert this loop into a Numpy expression.

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,339 @@
...
+        Nn = len(n)
+        IC1mu = np.empty(Nn, dtype=object)
+
+        for i in range(Nn):
+            gamma_i = 1 + em / mass_mev
+            beta_i = np.sqrt(1 - 1.0 / gamma_i**2)
+            fCi = np.maximum(0, 1 - 1.0 / n[i] ** 2 / beta_i**2)
+            intg_i = gEE * Fm * fCi
+            cs_val = cs_intg(1.0) * intg.unit
+            Lnorm = 2 * np.pi / 137 / rho / Lam[i].to(u.cm) ** 2 * cs_val
```
lpsinger, line 129: Is 137 here the hardcoded value of the fine structure constant?

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,339 @@
...
+            kind="integral",
+            solar=solar,
+            particle=particle,
+        )
+        for e in energy_bins
+    ]
+    return Table([energy_bins, u.Quantity(flux_integral)], names=["energy", "flux"])
+
+
+@dataclass
```
lpsinger, line 44: Why is this a class?

lpsinger (changes requested):

m4opt/synphot/background/_cerenkov/__init__.py
weizmannk: @lpsinger  The way I am calling `CerenkovEmission` and `CerenkovBackground` requires them to use the @dataclass decorator. Without @dataclass, those keyword argument initializations wouldn’t work.

weizmannk:
https://github.com/m4opt/m4opt/pull/334#discussion_r2109203239

@lpsinger  The way I am calling `CerenkovEmission` and `CerenkovBackground` requires them to use the `@dataclass` decorator. Without that, those keyword argument initializations wouldn’t work.

weizmannk:
@lpsinger Yes, you are right. I was doing this mainly to limit the energy range from 0.04 to 8 MeV, similar to Figure 6 of Kruk et al. (2016): https://iopscience.iop.org/article/10.1088/1538-3873/128/961/035005.
In other words, it was just to restrict the range. I will remove the extra interpolation step and simply use 1000 bins over the desired energy interval.

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+        ee = np.logspace(np.log10(0.04), np.log10(8.0), 1000) * u.MeV
+
+        # --- Interpolate the electron flux onto the energy grid
+        cs_flux = CubicSpline(
+            flux_data["energy"].value,
+            flux_data["flux"].value,
+            bc_type="natural",
+            extrapolate=True,
+        )
+        Fe = u.Quantity(cs_flux(ee.value), flux_data["flux"].unit)
```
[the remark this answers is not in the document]
weizmannk: @lpsinger Yes, you are right. I was doing this mainly to limit the energy range from 0.04 to 8 MeV, similar to Figure 6 of Kruk et al. (2016): https://iopscience.iop.org/article/10.1088/1538-3873/128/961/035005.
In other words, it was just to restrict the range. I will remove the extra interpolation step and simply use 1000 bins over the desired energy interval.

weizmannk:
@lpsinger It's done , https://github.com/m4opt/m4opt/pull/334#discussion_r2114954147

weizmannk:
@lpsinger I think this is also ready or close to being merge.

weizmannk:
@lpsinger  I have added `test_cerenkov.py` for the unit tests, and those are now passing. For now, I am just returning the original intensity as in [the MATLAB implementation](https://github.com/EranOfek/AstroPack/blob/main/matlab/astro/%2Bultrasat/Cerenkov.m#L217).

I guess that eventually we will want to return the wavelength and intensity as a synphot spectrum, right? If so, I’ll go ahead and add the correct conversion.
 to be uncomment after the test unit test pass:  `SourceSpectrum(Empirical1D, points=wavelength, lookup_table=intensity_photlam)`

lpsinger (changes requested):

lpsinger:
> @lpsinger I have added `test_cerenkov.py` for the unit tests, and those are now passing. For now, I am just returning the original intensity as in [the MATLAB implementation](https://github.com/EranOfek/AstroPack/blob/main/matlab/astro/%2Bultrasat/Cerenkov.m#L217).

These tests only check that the value has the correct units, not the correct value.

> I guess that eventually we will want to return the wavelength and intensity as a synphot spectrum, right? If so, I’ll go ahead and add the correct conversion: `SourceSpectrum(Empirical1D, points=wavelength, lookup_table=intensity_photlam)`

Yes.

weizmannk:
@lpsinger Yes, I just checked the units. I don’t have any reference values from the MATLAB output to check against. To do that, I would need to take some time since I have little or no experience with MATLAB.

lpsinger (changes requested):
The new comparison plot looks great, but we need an automated test that checks that the results are close that does _not_ rely on a person looking at the plot. Here are some options:

- Use [pytest-mpl](https://github.com/matplotlib/pytest-mpl) to create an image comparison test
- Add a test that uses [numpy.testing](https://numpy.org/doc/stable/reference/routines.testing.html) to compare the array output

weizmannk:
@lpsinger yes the pytest is done with : [numpy.testing(https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose)

Here is the test file : https://github.com/m4opt/m4opt/pull/334/commits/7ae3aed9777cb5ffa4de51ba9b713f371ca67991#diff-70685a29ff4b474bcdbbd5d1ab3a919a1e78201785a5481b8837a0600a5e374a

lpsinger:
> Okay, I see. I thought the test wasn’t covered by Sphinx.

Example plots in the documentation aren't included in the tests, but they are automatically generated and included in the documentation. There are lots of examples of generated plots in the documentation.

weizmannk:
@lpsinger Yes, the problem isn’t how to implement the .. plot:: directive, I know how to do that. I think we might not be on the same page, or I am a bit lost. For the test, I am reading the AE9 electron flux because that’s what is used in the Matlab code ([test here](https://github.com/weizmannk/m4opt/blob/cerenkov-background-light/m4opt/tests/cerenkov_test/test_cerenkov.py#L18)).
But in `CerenkovBackground`, as covered by the API, it reads the AE8 model. So I want to avoid importing the function from the test into CerenkovBackground just for plotting.
However, if you think it’s fine, I can import the function from the test and include the plot alongside the others in `api/m4opt.synphot.background.CerenkovBackground`.

lpsinger (changes requested):

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -288,20 +289,20 @@ class CerenkovBackground:
...
+        ax.loglog(Ek.value, dEdX_Al2O3.value, label="Sapphire (Al$_2$O$_3$)", linewidth=lw)
+
+        ax.set_xlabel("Electron kinetic energy [MeV]", fontsize=label_fs)
+        ax.set_ylabel(r"Energy loss dE/dX [MeV/(g cm$^{-2}$)]", fontsize=label_fs)
+        ax.set_title(r"Electron Energy Loss per Unit Mass Thickness", fontsize=label_fs)
+        ax.legend(fontsize=label_fs)
+        ax.grid(True, which='both', ls='--')
+        ax.tick_params(labelsize=tick_fs)
+        fig.tight_layout()
         plt.show()
```
lpsinger, line 306: Not necessary in plot directives.

```suggestion
```

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -314,17 +315,18 @@ class CerenkovBackground:
...
+        ax.set_xlabel(r"Electron kinetic energy [MeV]")
+        ax.set_ylabel(r"Inverse energy loss $(dE/dX)^{-1}$ [(g cm$^{-2}$)/MeV]")
+        ax.set_title(r"Inverse Energy Loss for Silica (SiO$_2$)")
+        ax.grid(True, which='both', ls='--')
+        ax.axis([1e-2, 1e2, 1e-2, 1])
+        ax.tick_params(labelsize=10)
+        fig.tight_layout()
         plt.show()
 
+
```
lpsinger, line 329:
```suggestion
```

lpsinger (changes requested):

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+    )
+    ee = flux_data["energy"]
+    Fe = flux_data["flux"]
+
+    # --- Material optical parameters: index of refraction and density
+    # Dict: material name -> (refractive index at ~1 micron, density)
+    material_properties = {
+        "sio2": (1.5, 2.2 * u.g / u.cm**3),
+        "SiO2_suprasil_2a": (1.5, 2.2 * u.g / u.cm**3),
+        "sapphire": (1.75, 4.0 * u.g / u.cm**3),  # at 1 mu, n(0.25 mu)=1.85
```
lpsinger, line 70: The comment says 1.85, but the numeric literal is 1.75.

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+    }
+    if material not in material_properties:
+        raise ValueError(f"Unknown material option: '{material}'")
+
+    n_val, rho = material_properties[material]
```
lpsinger, line 75:
```suggestion
    try:
        n_val, rho = material_properties[material]
    except KeyError:
        raise ValueError(f"Unknown material option: '{material}'")

```

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+
+    # Compute midpoint energies between grid bins
+    em = 0.5 * (ee[:-1] + ee[1:])
+
+    # Get particle mass (rest energy) in MeV
+    mass = m_e if particle == "e" else m_p
+    mass_mev = (mass * c**2).to("MeV")
```
lpsinger, line 82:
```suggestion
    mass_mev = (mass * c**2).to(u.MeV)
```

m4opt/tests/cerenkov_test/test_cerenkov.py
```diff
@@ -0,0 +1,400 @@
...
+    Reads a MATLAB struct-style XML file and returns the results as an Astropy Table.
+
+    The XML file contains the Cerenkov emission calculated using the daily flux (DailyMax)
+    at the 95th percentile, for the material: 'si02_suprasil_2a'.
+
+    Args:
+        xml_path (str): Path to the XML file.
+        fields (list or None): List of fields/tags to extract. If None, detects all automatically.
+
+    Returns:
```
lpsinger, line 133: Please use Numpydoc-style comments.

m4opt/tests/cerenkov_test/test_cerenkov.py
```diff
@@ -0,0 +1,400 @@
...
+        extrapolate=True,
+    )
+    Fe = u.Quantity(cs_flux(ee.value), flux.unit)
+
+    # Compute midpoint energies between grid bins
+    em = 0.5 * (ee[:-1] + ee[1:])
+
+    # Get particle mass (rest energy) in MeV
+    mass = m_e if particle == "e" else m_p
+    mass_mev = (mass * c**2).to("MeV")
```
lpsinger, line 243:
```suggestion
    mass_mev = (mass * c**2).to(u.MeV)
```

m4opt/tests/cerenkov_test/test_cerenkov.py
```diff
@@ -0,0 +1,400 @@
...
+
+    # Truncate to minimum length (in case of unequal arrays)
+    min_len = min(len(lst) for lst in data.values())
+    data = {k: v[:min_len] for k, v in data.items()}
+
+    # Create and return Astropy Table
+    return Table(data)
+
+
+def cerenkov_emission_py(
```
lpsinger, line 181: A lot of this looks like it is copy-pasted from your code in `m4opt.synphot.background._cerenkov`. Why? It's not a valid test if you are merely testing the code against a copy of itself.

m4opt/tests/cerenkov_test/__init__.py
lpsinger, line 1: Please move this entire directory to `m4opt/synphot/background/_cerenkov/tests`.

m4opt/tests/cerenkov_test/test_cerenkov.py
```diff
@@ -0,0 +1,400 @@
...
+import numpy as np
+import pytest
+from astropy import units as u
+from astropy.constants import alpha, c, m_e, m_p
+from astropy.table import Table
+from scipy.interpolate import CubicSpline
+
+from m4opt.synphot.background._cerenkov._electron_loss import get_electron_energy_loss
+from m4opt.synphot.background._cerenkov._refraction_index import get_refraction_index
+from m4opt.synphot.background._core import BACKGROUND_SOLID_ANGLE
```
lpsinger, line 15: Use relative imports.

lpsinger (changes requested):

m4opt/synphot/background/_cerenkov/tests/test_cerenkov.py
lpsinger, line 1: There is still a lot of code in here that looks like it is a _copy_ of the code under test. Please test the actual module, not a copy of it.

lpsinger (changes requested):

weizmannk:
@lpsinger As I mentioned earlier, importing a module would not be very relevant in this context.
The main goal is to ensure that the translation has been done correctly.
I have now separated the reading and calculation of AE8, which allows for comparing or supplementing another flux before applying the Cerenkov emission.
Although the process is correct, it currently seems somewhat difficult to read.

Nevertheless, this approach ensures the proper use of module importation for comparison purposes.

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,370 @@
...
+
+import numpy as np
+from aep8 import flux
+from astropy import units as u
+from astropy.constants import alpha, c, m_e, m_p
+from astropy.table import Table
+from scipy.interpolate import CubicSpline
+from synphot import Empirical1D, SourceSpectrum
+
+# from synphot import Empirical1D, SourceSpectrum
```
lpsinger, line 12: Remove commented out code.

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,370 @@
...
+        :include-source: False
+        :caption: Cerenkov background spectrum for GEO at 2025-05-18T02:48:00Z
+
+        from astropy.coordinates import EarthLocation, ICRS
+        from astropy_healpix import HEALPix
+        from astropy.time import Time
+        from astropy import units as u
+        import numpy as np
+        import matplotlib.pyplot as plt
+        plt.rcParams["font.family"] = "Times New Roman"
```
lpsinger, line 231: Please do not change the font.

```suggestion
```

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,370 @@
...
+        with observing(observer_location=observer_location, target_coord=coord, obstime=obstime):
+            cerenkov_model = CerenkovBackground(particle='e', solar='max', material='SiO2_suprasil_2a')
+            spectrum = cerenkov_model.cerenkov_emission(observer_location, obstime)
+            wavelength = spectrum.waveset
+            intensity = spectrum(wavelength)
+
+        fig, ax = plt.subplots(figsize=(7, 5))
+        ax.plot(wavelength, intensity)
+        ax.set_xlabel(rf"Wavelength [{wavelength.unit}]")
+        ax.set_ylabel(rf"Intensity [{intensity.unit} (erg / s cm$^{{2}}$ Hz)]")
```
lpsinger, line 252: This is probably redundant because `{intensity.unit}` probably _renders to_ `erg / s cm$^{{2}}$ Hz)`.

m4opt/synphot/background/_cerenkov/__init__.py
```diff
@@ -0,0 +1,370 @@
...
+        ----------
+        observer_location : EarthLocation
+            Observer location, e.g., in geodetic coordinates.
+        obstime : Time
+            Observation time.
+
+        Returns
+        -------
+        SourceSpectrum
+            The computed Cerenkov background spectrum as a synphot SourceSpectrum.
```
lpsinger, line 360: Use type annotations, not docstring comments, to indicate types.

Also, the observer location does _not_ need to be in geodetic coordinates; this will work whether it is in geographic or geodetic coordinates.

```suggestion
    def cerenkov_emission(self, observer_location: EarthLocation, obstime: Time) -> SourceSpectrum:
        """
        Compute the Cerenkov emission spectrum for the given observer and time.

        Parameters
        ----------
        observer_location
            Observer location.
        obstime
            Observation time.

        Returns
        -------
        :
            The computed Cerenkov background spectrum as a synphot SourceSpectrum.
```

m4opt/synphot/background/_cerenkov/tests/test_cerenkov.py
```diff
@@ -0,0 +1,310 @@
...
+    intensity_py = spectrum_py(wavelength_py)
+
+    # Enforce the check for python output
+    # Check that the photlam intensity is compatible with the one wait for  synphot Spectrum
+    # https://github.com/EranOfek/AstroPack/blob/main/matlab/astro/%2Bultrasat/Cerenkov.m#L217
+
+    # Check value accuracy with MATLAB reference
+    assert np.allclose(wavelength_mat.value, wavelength_py.value, rtol=1e-10), (
+        "Wavelength mismatch"
+    )
```
lpsinger, line 277:
```suggestion
    np.testing.assert_allclose(wavelength_mat.value, wavelength_py.value, rtol=1e-10), err_msg="Wavelength mismatch")
```

m4opt/synphot/background/_cerenkov/tests/test_cerenkov.py
```diff
@@ -0,0 +1,310 @@
...
+    rel_diff = np.abs(intensity_mat.value - intensity_py.value) / (
+        np.abs(intensity_mat.value) + 1e-30
+    )
+    assert np.all(rel_diff < 4e-2), (
+        f"Python vs MATLAB: maximum relative difference {np.max(rel_diff):.2e} exceeded"
+    )
```
lpsinger, line 283: Use similar `numpy.testing.assert_...` method here.

Why do you need to calculate the relative error yourself, instead of passing a relative tolerance to the assertion?

lpsinger (changes requested):

PR #336, opened 2025-05-19
Add all-sky grid

weizmannk:
This PR adds the **All-Sky Survey (AllSS)** grid for **ULTRASAT**. During the first 6 months, ULTRASAT will use the grid intended for the All-Sky Survey (AllSS) instead of the **Low-Cadence Survey (LCS)**.

The **LCS grid** requires no overlap between fields, which are defined as circles with 7° radii.  
The **AllSS grid**, however, ensures that no point on the sky is farther than 7° from a given grid point, allowing for overlap between fields.

According to Yossi, the LCS grid is not the relevant one for this purpose, but we will compare both the non-overlapping LCS grid and the All-Sky grid for future evaluations.

We want this label to be selected using a  `--skygrid`  flag by default, the `--skygrid` is` None` and the first skygrid is read , in this case  **AllSS** one. 

Notes
=====

- A `--skygrid` flag is added to allow selecting the desired grid.  
- If `--skygrid` is **not specified** (i.e., `None`), for ULTRASAT we need to specify the select it.   
- For ULTRASAT, `Mission.skygrid` is of type:   `dict[Hashable, SkyCoord]`  
eg: `{"allsky": SkyCoord(...),  "non-overlap": SkyCoord(...)}`

By now the other missions, nothing changes:  `Mission.skygrid` remains of type` SkyCoord`.


lpsinger (commented):
Here's my suggestion. Change the type of `Mission.skygrid` to `SkyCoord | dict[Hashable, SkyCoord]` so that a mission may have a dictionary of sky grids. Then add a `--skygrid` option to the `m4opt plan` command line interface to select the sky grid.

m4opt/missions/_ultrasat/__init__.py
```diff
@@ -20,11 +20,24 @@
...
+        )
+        ra = [float(str(s).strip().rstrip(",")) for s in table["col1"]]
+        dec = table["col2"]
+
+    else:
+        table = Table.read(
+            resources.files(data) / "LCS_nonoverlapping_grid.csv", format="ascii.csv"
+        )
+        ra = table["RA"]
+        dec = table["Dec"]
```
lpsinger, line 38: These two branches are so dissimilar that it would be better as two separate functions, rather than one function with an argument.

m4opt/missions/_ultrasat/data/README.rst
```diff
@@ -1,9 +1,12 @@
...
-The ULTRASAT sky grid includes 240 non-overlapping fields (7 radius) covering the entire sky
-for the low-cadence extragalactic survey. Each field is annotated with continuous visibility
-for at least one 180-day or 45-day period per year, and average UV extinction.
+ULTRASAT includes two predefined sky grids:
 
-The grid is based on the baseline survey strategy discussed within the ULTRASAT Working Groups
+- **LCS grid** (default): 240 non-overlapping fields (7° radius) covering the entire sky, used for the low-cadence extragalactic survey.
+  Each field includes visibility annotations (180-day or 45-day periods) and average UV extinction.
+
+- **ALLSS grid**: Used during the first 6 months of the mission. Overlapping 7° fields ensure no point on the sky is farther than 7° from a grid center.
```
lpsinger, line 9: List this one first, because it is active earlier in the mission.

m4opt/missions/_ultrasat/data/README.rst
lpsinger, line 1: Please define the acronyms LCS and ALLSS.

m4opt/missions/_ultrasat/data/AllSS_grid_361.txt
```diff
@@ -0,0 +1,361 @@
...
+151, -26
+94.5, -8
+108.5, -11.2
+265.8, -14
+314.8, -12
+50.5, 9.5
+349.9, 9.4
+341.7, 32.7
+254, 39
+50, 61.5
```
lpsinger, line 361: Why do these numbers have a different format?

m4opt/_cli/schedule.py
```diff
@@ -275,7 +282,20 @@ def schedule(
...
+            if skygrid is not None:
+                if skygrid not in mission.skygrid:
+                    typer.secho(
+                        f"skygrid '{skygrid}' not found. Options: {', '.join(mission.skygrid.keys())}",
+                        fg=typer.colors.RED,
+                    )
+                    raise typer.Exit(1)
+                target_coords = mission.skygrid[skygrid]
+            else:
+                target_coords = next(iter(mission.skygrid.values()))
```
lpsinger, line 292: Maybe it would be better to have no default, and error out if there is more than one sky grid and none are specified.

m4opt/_cli/schedule.py
```diff
@@ -275,7 +282,20 @@ def schedule(
...
+        if isinstance(mission.skygrid, dict):
+            if skygrid is not None:
+                if skygrid not in mission.skygrid:
+                    typer.secho(
+                        f"skygrid '{skygrid}' not found. Options: {', '.join(mission.skygrid.keys())}",
+                        fg=typer.colors.RED,
+                    )
+                    raise typer.Exit(1)
+                target_coords = mission.skygrid[skygrid]
+            else:
```
lpsinger: This can be rewritten with fewer nested branches:

```python
if not isinstance(mission.skygrid, dict):
    target_coord = mission.skygrid
elif skygrid in mission.skygrid:
    target_coord = mission.skygrid[skygrid]
else:
    ... # print error, exit
```

m4opt/_cli/schedule.py
lpsinger, line 1: Also update animate.py, which should pull the value of `skygrid` from the table metadata.

weizmannk:
@lpsinger  It’s done. I have added your suggestion, and now the skygrid can be specified using the **--skygrid** option.

m4opt/_cli/schedule.py
```diff
@@ -129,6 +129,13 @@ def schedule(
     mission: Annotated[
         missions.Mission, typer.Option(show_default="uvex")
     ] = missions.uvex,
+    skygrid: Annotated[
+        str | None,
+        typer.Option(
+            help="Name of sky grid to use, e.g. 'overlap' or 'allsky'.",
```
lpsinger:
```suggestion
            help="Name of sky grid to use, if the mission supports multiple sky grids.",
```

lpsinger (changes requested):

m4opt/_cli/schedule.py
weizmannk: @lpsinger Do you mean that I should add the `skygrid` key? The value is already saved in the metadata, https://github.com/m4opt/m4opt/blob/main/m4opt/_cli/schedule.py#L652. Maybe I missed something?

weizmannk: Ah! Forget about it, I see now. I forgot that here the value is the file name.

weizmannk:
@lpsinger Due to the fact that we use` with status()` to provide progress descriptions, any print or` typer.secho()` statements inside that block are captured or suppressed. As a result, when the `—skygrid` option is incorrect, the following code does not display any output in the terminal:
```
typer.secho(
    f"skygrid '{skygrid}' not found. Options: {', '.join(map(str, mission.skygrid.keys()))}",
    fg=typer.colors.RED,
)
```

m4opt/_cli/schedule.py
lpsinger: Oh, right. I guess that `animate.py` doesn't need to know about the sky grid. But please make sure that `schedule.py` does _save_ the value of the argument in `plan_args`.

m4opt/_cli/schedule.py
```diff
@@ -275,7 +281,17 @@ def schedule(
...
+        if not isinstance(mission.skygrid, dict):
+            target_coords = mission.skygrid
+        elif skygrid in mission.skygrid:
+            target_coords = mission.skygrid[skygrid]
+        else:
+            typer.secho(
+                f"skygrid '{skygrid}' not found. Options: {', '.join(map(str, mission.skygrid.keys()))}",
+                fg=typer.colors.RED,
+            )
+            raise typer.Exit(1)
```
lpsinger, line 293:
```suggestion
            raise UsageError(
                f"skygrid '{skygrid}' not found. Options: {', '.join(map(str, mission.skygrid.keys()))}"
            )
```

And add `from click import UsageError`.

For uncaught exceptions that are instances of any of the [Click exception classes](https://click.palletsprojects.com/en/stable/api/#exceptions), typer will nicely format and print the exception message but hide the traceback.

lpsinger (changes requested):

lpsinger:
> @lpsinger Due to the fact that we use` with status()` to provide progress descriptions, any print or` typer.secho()` statements inside that block are captured or suppressed. As a result, when the `—skygrid` option is incorrect, the following code does not display any output in the terminal:
> 
> ```
> typer.secho(
>     f"skygrid '{skygrid}' not found. Options: {', '.join(map(str, mission.skygrid.keys()))}",
>     fg=typer.colors.RED,
> )
> ```

See https://github.com/m4opt/m4opt/pull/336#discussion_r2103450955.

weizmannk:
@lpsinger Thanks, everything is working fine now!

m4opt/_cli/schedule.py
```diff
@@ -616,7 +631,8 @@ def schedule(
         with status("writing results"):
             if write_progress is not None:
                 QTable(
-                    rows=recorder.recorded,
+                    # FIXME: workaround for https://github.com/astropy/astropy/issues/17688
+                    rows=recorder.recorded if recorder.number_of_records > 0 else None,
```
lpsinger, line 635: Please undo. Looks like a bad merge.

m4opt/missions/_ultrasat/data/AllSS_grid_361.txt
lpsinger: Please respond to this comment.

m4opt/missions/_ultrasat/__init__.py
```diff
@@ -20,7 +20,18 @@
...
 
-def _read_skygrid():
+def _read_allsky_skygrid() -> SkyCoord:
+    # Load the All-Sky Survey (AllSS) grid.
+    table = Table.read(
+        resources.files(data) / "AllSS_grid_361.txt", format="ascii.no_header"
+    )
+    ra = [float(str(s).strip().rstrip(",")) for s in table["col1"]]
+    dec = table["col2"]
+    return SkyCoord(ra, dec, unit=u.deg)
```
lpsinger, line 30:
```suggestion
    table = Table.read(
        resources.files(data) / "AllSS_grid_361.txt", format="ascii.csv", data_start=0, names=["ra", "dec"]
    )
    return SkyCoord(table["ra"], table["dec"], unit=u.deg)
```

lpsinger (changes requested):

lpsinger:
@weizmannk, this looks good, except for the following two issues. Please address them:

- https://github.com/m4opt/m4opt/pull/336#discussion_r2104300635
- https://github.com/m4opt/m4opt/pull/336#discussion_r2104297866

weizmannk:
@lpsinger I may have missed or misunderstood your message.
-https://github.com/m4opt/m4opt/pull/336#discussion_r2098634928

By "the format," do you mean the fact that the RA values are in string format while the DEC values are floats?
If so, here’s the response:

> We received them this way, and I’m not sure why the RA is in string format while the DEC is a float. That’s why I converted the string to a float first. @mcoughlin , do you happen to know why the RA is in string format and the DEC is in float format in the allsky grid ".txt" file?

For the other issue, I failed to undo it correctly, so I reverted the relevant lines instead:
https://github.com/m4opt/m4opt/pull/338

lpsinger (approved):

PR #347, opened 2025-05-30
Overload logical operators on constraints

lpsinger:
This simplifies evaluating a mission's constraints because you don't have to loop over them yourself.


mcoughlin (approved):
Agreed that this is more elegant.

PR #348, opened 2025-05-30
Add wrapper function to partition a graph with METIS

lpsinger:


lpsinger:
@inducer, thanks for maintaining the very nice Python wrapper, pymetis! You may be interested in the code here which completes the integration with NetworkX.

mcoughlin (approved):
A few ideas for docs.

PR #355, opened 2025-06-05
Temporarily remove untested LSST ETC model

lpsinger:


mcoughlin (approved):
@weizmannk just tagging so you know

PR #434, opened 2026-02-20
ZTF detector model

mcoughlin:
This PR adds the ZTF instrument model and associated tests.

Closes #375

PR #435, opened 2026-02-20
Rubin instrument model

mcoughlin:
This PR adds a Rubin instrument model and tests.

### DIFF
--- a/m4opt/missions/_rubin/__init__.py
+++ b/m4opt/missions/_rubin/__init__.py
@@ -72,6 +72,7 @@ rubin = Mission(
         # "gain can safely be assumed to be 1" for SNR calculations
         gain=1,
         # Dark current requirement: 0.2 e-/s/pixel
+        # Table 4 of https://smtn-002.lsst.io/
         dark_noise=0.2 * u.Hz,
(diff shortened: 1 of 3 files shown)


m4opt/missions/_rubin/__init__.py
```diff
@@ -68,7 +85,55 @@ def _make_fov():
 focal plane consists of 189 detectors arranged in 21 rafts, as shown in Figure
 12 from :footcite:`2019ApJ...873..111I`.
 
+Note
+----
+Detector parameters are from `SMTN-002 <https://smtn-002.lsst.io/>`_
```
lpsinger: That is actually a Sphinx documentation set. Please use an Intersphinx reference.

lpsinger (changes requested):

lpsinger:
I will try to review these two PRs thoroughly today, or by Monday at the latest.

lpsinger (commented):
This PR is so similar to #434, that I think it will make the review faster if we finish iterating on and merge #434, and then you make the analogous changes to this PR.

lpsinger (approved):

PR #439, opened 2026-02-21
Earthshine background

mcoughlin:
This PR adds Earthshine as a background. 

Note that https://github.com/m4opt/m4opt/pull/438 should go in first and then rebase, as I was working on top of it.


mcoughlin:
@lpsinger Of course.

m4opt/synphot/background/_earthshine/tests/test_earthshine.py
```diff
@@ -0,0 +1,50 @@
...
+    loc = EarthLocation.from_geodetic(
+        lon=15 * u.deg, lat=0 * u.deg, height=35786 * u.km
+    )
+    coord = SkyCoord(0 * u.deg, 0 * u.deg)
+    obstime = Time("2025-05-18T02:48:00Z")
+
+    bg = EarthshineBackground()
+    with observing(observer_location=loc, target_coord=coord, obstime=obstime):
+        val = bg(2600 * u.AA)
+    assert val.value > 0
```
lpsinger: This is not a very constraining test. Would you please add a FIXME comment?

lpsinger (changes requested):

lpsinger (commented):
This is probably also spatially dependent, isn't it? How does it vary spatially?

lpsinger (changes requested):
What is the range of valid values of earth limb angle? Does this function have correct behavior in the extreme limits?

The earthshine must also depend on whether we are looking at the illuminated part of the Earth limb. Can we model that?

mcoughlin:
@lpsinger 

Limb angle: it spans about −9° to 171° at GEO. The scale factor is 0 below the limb and above 50° extrapolates on the slope from STIS, so should be good.

On illumination: I added the observer's position, so now it actually goes to zero over the night side, so that is good.

src/m4opt/synphot/background/_earthshine/tests/test_earthshine.py
```diff
@@ -0,0 +1,187 @@
...
+    sf = EarthshineBackgroundScaleFactor()
+
+    # Observer at 2*R_earth from center (altitude = R_earth above surface).
+    # limb_alt = arccos(R_earth / (2*R_earth)) = 60 deg.
+    # limb_angle = alt + 60 deg, so alt = limb_angle - 60 deg.
+    # At the North Pole, alt = dec.
+    # Use RA near the Sun so illumination factor ~ 1.
+    loc = EarthLocation.from_geocentric(0 * u.m, 0 * u.m, 2 * R_earth)
+
+    for limb_angle_deg, expected_scale in [(24, 2.0), (38, 1.0), (50, 0.5)]:
```
lpsinger: Use `pytest.mark.parametrize` instead of looping.

src/m4opt/synphot/background/_earthshine/tests/test_earthshine.py
```diff
@@ -0,0 +1,187 @@
...
+    # and the illumination variation can break strict monotonicity).
+    limb_angles = [24, 30, 38, 50, 60, 80, 100, 120]
+    scales = []
+    for la in limb_angles:
+        alt_deg = la - 60
+        coord = SkyCoord(_SUN_RA_DEG * u.deg, alt_deg * u.deg)
+        scales.append(sf.at(loc, coord, _TEST_OBSTIME))
+
+    for i in range(1, len(scales)):
+        assert scales[i] <= scales[i - 1]
```
lpsinger: No need to loop explicitly; all of these things vectorize.

src/m4opt/synphot/background/_earthshine/tests/test_earthshine.py
```diff
@@ -0,0 +1,187 @@
...
+    """Scale factor is monotonically decreasing with increasing limb angle."""
+    sf = EarthshineBackgroundScaleFactor()
+
+    loc = EarthLocation.from_geocentric(0 * u.m, 0 * u.m, 2 * R_earth)
+
+    # Test limb angles from 24 deg to 120 deg.
+    # Use RA near the Sun so illumination is roughly constant.
+    # Start from 24 deg (below that, the limb component is clamped at 2.0
+    # and the illumination variation can break strict monotonicity).
+    limb_angles = [24, 30, 38, 50, 60, 80, 100, 120]
```
lpsinger: What's with this list? What makes these values special?

lpsinger (changes requested):
I'll get to this in detail within a day or two. In the meanwhile it might be nice (but not required) to add a diagram (perhaps generated by Matplotlib, perhaps created in a drawing program) to illustrate the angles involved. Here's a diagram that I made for the roll angle, if it gives you some idea of the level of detail.

[roll-angle-flip.pdf](https://github.com/user-attachments/files/31666321/roll-angle-flip.pdf)

mcoughlin:
@lpsinger I can't see your attachment :(

src/m4opt/synphot/background/_earthshine/tests/test_earthshine.py
mcoughlin: I will just make these arange.

src/m4opt/synphot/background/_earthshine/tests/test_earthshine.py
mcoughlin: Fixed.

src/m4opt/synphot/background/_earthshine/tests/test_earthshine.py
mcoughlin: Yup.

lpsinger:
> @lpsinger I can't see your attachment :(

roll-angle-flip

mcoughlin:
geometry

@lpsinger maybe something like this?

lpsinger:
> @lpsinger maybe something like this?

Looks great. Did you just make that? Is this Illustrator, Keynote, LaTeX+TikZ, something else?

mcoughlin:
@lpsinger felipe and I had put together something related when discussing this exact thing!

lpsinger:
> @lpsinger felipe and I had put together something related when discussing this exact thing!

A few suggestions:

- As drawn, it looks like the direction to the target is parallel to the Sun's rays. Change the angle so that they are clearly oblique.
- Put the gamma label nearer the angle it is actually depicting. Don't point to it, just put the label near the angle. I tend to go for a dimensioning and labeling style that is close to what you would use in a mechanical drawing.
- Shade the sunlit and dark side.
- Make it a vector graphic.

I'm happy to take a stab at this in Illustrator if it's useful.

mcoughlin:
@lpsinger that would be great. This was just matplotlib so no doubt you can make it nicer that way!

lpsinger:
> @lpsinger that would be great. This was just matplotlib so no doubt you can make it nicer that way!

Well, you probably know my system, and that I often hop back and forth between Matplotlib and Illustrator. May I have your Matplotlib code?

mcoughlin:
[geometry_figure.py](https://github.com/user-attachments/files/31667518/geometry_figure.py)
@lpsinger here!

lpsinger (changes requested):
The docs have some glitches.

Image

mcoughlin:
@lpsinger I think I got these.

lpsinger:
I thought a good test of this would be to make a picture of the earthshine from a specific location in Earth orbit, looking directly back toward the Earth. It doesn't look like what I would expect. It isn't even centered on the Earth. But my code could be wrong; I did this very hastily.

```python
import warnings
warnings.filterwarnings("ignore", "Wswiglal-redir-stdio")

from matplotlib import pyplot as plt
import numpy as np
from ligo.skymap.plot import earth, sun
from astropy import units as u
from astropy.time import Time
from m4opt.synphot.background._earthshine import EarthshineBackgroundScaleFactor
from astropy.coordinates import EarthLocation, SphericalRepresentation, GCRS, get_body

altitude = 5 * u.Rearth
scale_factor = EarthshineBackgroundScaleFactor()
observer_location = EarthLocation(*SphericalRepresentation(0 * u.deg, 0 * u.deg, altitude).to_cartesian().xyz)
obstime = Time('J2000').utc
earth_coord = get_body('earth', obstime, observer_location)
sun_coord = get_body('sun', obstime, observer_location)
ax = plt.axes(projection='astro zoom', center=earth_coord, radius=100 * u.deg)
x = np.arange(*ax.get_xlim())
y = np.arange(*ax.get_ylim())
target_coord = ax.wcs.pixel_to_world(*np.meshgrid(x, y)).transform_to(earth_coord.frame)
ax.imshow(scale_factor.at(observer_location, target_coord, obstime))
ax.plot_coord(earth_coord, marker=earth, markersize=20, markeredgewidth=2)
ax.plot_coord(sun_coord, marker=sun, markersize=20, markeredgewidth=2)
plt.savefig('earth_image.png')
```
earth_image

mcoughlin:
@lpsinger That was a great find! So I think this is a frame artifact, as get_body('earth', obstime, observer_location) is in a topocentric frame, while the WCS grid is distance-less directions. I think things are "correct" in one frame, as then the dark disk lands on the limb circle with the zero at arcsin(R⊕/r) = 11.537°.

We get a result where the model gives a lopsided bowtie with a bunch of the sky outside the disk pinned at exactly zero when the earthshine should be a near-symmetric halo, so that doesn't seem good, and I think is an artifact of the grazing-point construction. 

So I think we can instead integrate incidence around the visible limb weighted by proximity to the line of sight so we won't get this singularity problem.

What do you think?

lpsinger:
The mistake is in my code. Let me fix it; I think this will be a useful visualization for the docs and also a good sanity check.

mcoughlin:
@lpsinger just pushed a fix to try to remove the artifact.

lpsinger:
Here's the corrected script and the output for the image of the Earthshine as a function of the observer location. Would you please add this as a `.. plot:` block in the documentation?

[earthshine-gallery.py](https://github.com/user-attachments/files/31751867/earthshine-gallery.py)

earthshine-gallery

In this script as I changed the distance from 5 Rsun to thousands of Rsun, the angular diameter of the occulted area shrunk down like I expected but the Earthshine itself did not.

I suggest adding tests of:

- observers at various distances from the Earth, check that the angular diameter shrinks down as you get further out
- document the maximum distance for which this code is correct. For example, consider the arrangement of the sun, earth, and observer in conjunction, like this:

sun --- earth --- observer

When the observer is at a distance of $(R_\odot / R_\oplus - 1)^{-1} au \approx 217 R_\oplus$, the angular diameter of the Earth and Sun are equal. Does the code give the correct values around and beyond this transitional distance? If not, then is there a maximum distance for which this code is correct? If so then document.

src/m4opt/synphot/background/_earthshine/__init__.py
```diff
@@ -0,0 +1,226 @@
...
+    of sight is tangent to the Earth, an angle :math:`\\arccos(R_\\oplus / r)`
+    from the sub-observer point. The cosine of the solar incidence angle is
+    averaged around that circle, weighting each point by how close it lies to
+    the line of sight, on the angular scale of the Earth itself.
+
+    Averaging rather than taking the single nearest point matters for an
+    observer far from the Earth, where the whole limb approaches the terminator
+    and the incidence angle at any one point of it is a poor stand-in for how
+    brightly the Earth shines into the telescope.
+    """
```
lpsinger: Please move this explanation to a docstring on one of the symbols that is exported so that it is rendered in the docs. For example, on `EarthshineBackground` itself.

lpsinger (changes requested):

src/m4opt/synphot/background/_earthshine/__init__.py
```diff
@@ -0,0 +1,304 @@
...
+                    vmax=2,
+                    origin="lower",
+                    extent=extent,
+                )
+        for ax, lon in zip(axs[-1], lons):
+            ax.set_xlabel(f"{lon:latex}")
+
+        fig.supxlabel("Geocentric longitude of observer")
+        fig.supylabel("Geocentric latitude of observer")
+        fig.colorbar(image, ax=axs, label="Earthshine scale factor")
```
lpsinger: You didn't like my little arrows showing the direction to the sun?

I thought it was a nice check that the earthshine was oriented correctly. And it looked cool.

src/m4opt/missions/_ultrasat/__init__.py
```diff
@@ -68,10 +69,13 @@ def _read_nonoverlapping_skygrid() -> SkyCoord:
...
         bandpasses={"NUV": _read_throughput()},
-        # FIXME: Add model for stray light
+        # The earthshine model is calibrated against HST in low Earth orbit, so
+        # it is renormalized here to the worst-case stray light of 12 e-/pix
+        # per 300 s in the ULTRASAT noise budget.
         background=GalacticBackground()
         + ZodiacalBackground()
-        + CerenkovBackground(factor=21),
+        + CerenkovBackground(factor=21)
+        + EarthshineBackground(factor=20),
```
lpsinger: Where does this number come from?

src/m4opt/synphot/background/_earthshine/__init__.py
```diff
@@ -0,0 +1,304 @@
...
+    Constant "high" earthshine spectrum (no spatial dependence):
+
+    >>> from astropy import units as u
+    >>> from m4opt.synphot.background import EarthshineBackground
+    >>> background = EarthshineBackground.high()
+    >>> float(background(5000 * u.angstrom).value) > 0
+    True
+
+    .. plot::
+        :caption: Earthshine seen from a grid of observer positions, all at
```
lpsinger, line 197:
```suggestion
        :include-source: False
        :caption: Earthshine seen from a grid of observer positions, all at
```

Also move this up and out of the example section. I wanted to include this to aid in understanding and to serve as a visual check of correctness. We shouldn't show the source code in the example section because this code is too complex to serve as a good example.

src/m4opt/synphot/background/_earthshine/__init__.py
```diff
@@ -0,0 +1,304 @@
...
+    by a factor that depends on the angular distance between the target and
+    the limb, and on the solar illumination of the limb.
+
+    The dependence on limb angle comes from the STIS Instrument Handbook and
+    the STScI Exposure Time Calculator, which give earthshine levels at three
+    limb angles:
+
+    - 24 degrees from the limb: 2.0x the "high" spectrum ("extremely high")
+    - 38 degrees from the limb: 1.0x the "high" spectrum ("high", baseline)
+    - 50 degrees from the limb: 0.5x the "high" spectrum ("average")
```
lpsinger, line 180: Ah, this is the angle of the line of sight from the Earth limb? That's why it doesn't change angular diameter correctly as you increase in distance from the Earth.

HST orbits at an altitude of less than 0.1 Earth radii, whereas ULTRASAT will orbit at an altitude of more than 5 Earth radii. I think that the shrinking of angular diameter with distance is a necessary effect to model.

lpsinger (changes requested):

src/m4opt/synphot/background/_earthshine/__init__.py
mcoughlin: Sorry I was copying stuff around. Should be back where it belongs now!

src/m4opt/missions/_ultrasat/__init__.py
mcoughlin: This was to match the ULTRASAT paper 12 e-, but I don't think this makes sense any longer now that we do the angular scaling properly (and this gives 0 e-)...

src/m4opt/synphot/background/_earthshine/__init__.py
mcoughlin: fixed.

src/m4opt/synphot/background/_earthshine/__init__.py
```diff
@@ -279,12 +261,100 @@ class EarthshineBackground:
...
+    absolute normalization should be trusted. Use the ``factor`` argument to
+    renormalize the model to an observatory's own stray light budget.
+
+    Three separate things limit how far from the Earth this model may be used.
+    The geometry of the limb itself is exact at any distance, but:
+
+    - The scale factor is a function of the angle from the limb alone, in absolute
+      degrees, so it carries no dependence on the observer's distance. The
+      earthshine of a distant observer does not fall off as the inverse square of
+      that distance the way it should, and ``factor`` has to absorb the difference.
```
lpsinger: But you do include the distance now.

src/m4opt/synphot/background/_earthshine/__init__.py
mcoughlin: I added this in. I now get 0 e-. The paper says this comes from their Baffle modeling, so maybe we just believe them?

lpsinger (changes requested):

src/m4opt/missions/_ultrasat/__init__.py
lpsinger: Is there a value of this factor that matches the ULTRASAT paper numbers?

lpsinger: I don't know that we do it _correctly_, but we do it better. HST is observing from firmly inside the exosphere whereas maybe ULTRASAT will be looking down at the exosphere from above? I think that this is a deep rabbit hole that we could fall into and what you have is pretty good already if we can make it match up with the ULTRASAT number.

mcoughlin: 2.5×10⁹ ...

src/m4opt/synphot/background/_earthshine/__init__.py
mcoughlin: yeah stale.

src/m4opt/missions/_ultrasat/__init__.py
lpsinger: Ha! Can you point me to the paper where they calculate this 12e-?

mcoughlin: https://arxiv.org/abs/2304.14482 Table 3.

lpsinger: Do you have a code snippet for calculating the number of e-?

mcoughlin:
```
import numpy as np
import synphot
from astropy import units as u
from astropy.coordinates import SkyCoord
from astropy.time import Time
from dataclasses import replace

from m4opt.constraints._earth_limb import _get_angle_from_earth_limb
from m4opt.missions import ultrasat
from m4opt.synphot import observing
from m4opt.synphot.background import (
    CerenkovBackground,
    EarthshineBackground,
    GalacticBackground,
    ZodiacalBackground,
)

detector = ultrasat.detector
exptime = 300 * u.s
flat = synphot.SourceSpectrum(synphot.ConstFlux1D, amplitude=0 * u.ABmag)


def electrons_per_pixel(background, observer_location, target_coord, obstime):
    """Background in e-/pix, the same path Detector.get_limmag takes."""
    with observing(observer_location, target_coord, obstime):
        _, rate = replace(detector, background=background)._get_count_rates(flat, "NUV")
    return (rate * exptime).to_value(u.dimensionless_unscaled)


# Table 3 quotes the zodiacal light for a field near the ecliptic pole
# viewed perpendicular to the Sun.
obstime = Time("2027-03-21T00:00:00")
observer_location = ultrasat.observer_location(obstime)
pole = SkyCoord(0 * u.deg, 90 * u.deg, frame="barycentricmeanecliptic").icrs

for name, background in [
    ("zodiacal", ZodiacalBackground()),
    ("galactic", GalacticBackground()),
    ("cerenkov", CerenkovBackground(factor=21)),
]:
    print(f"{name:10s} {electrons_per_pixel(background, observer_location, pole, obstime):8.2f}")

# Stray light is quoted as a maximum, so search the field of regard for it.
ra, dec = np.meshgrid(np.arange(0, 360, 3), np.arange(-89, 90, 3))
grid = SkyCoord(ra.ravel() * u.deg, dec.ravel() * u.deg)
worst = 0.0
for iso in ["2027-03-21", "2027-06-21", "2027-09-21", "2027-12-21"]:
    obstime = Time(iso)
    observer_location = ultrasat.observer_location(obstime)
    limb = _get_angle_from_earth_limb(observer_location, grid, obstime).to_value(u.deg)
    counts = electrons_per_pixel(EarthshineBackground(), observer_location, grid, obstime)
    worst = max(worst, counts[limb >= 48].max())
print(f"earthshine {worst:8.2e}  worst case at the 48 deg limb constraint")
print(f"           {12 / worst:8.2e}  factor needed for the 12 e- in Table 3")
```

lpsinger: Was my intuition that the apparent angular diameter of the earthshine should decrease with distance from the Earth, wrong?

I was thinking that the Earthshine was due to sunlight scattered by the extended exosphere, but I now think that's wrong. I think it's just sunlight _reflected_ from the Earth's surface/atmosphere/clouds/whatever as a surface, based on the ULTRASAT paper's description of estimating it from the Earth's albedo and the solar irradiance.

In that case the apparent angular diameter _should not_ change much with distance, right? The apparent angular diameter should depend primarily upon the off-axis light rejection properties of the telescope.

mcoughlin: Yeah I was thinking about that too. For stray light, the detector background is the source brightness convolved with the instrument's off-axis scattering function.  Since we are far away from the Earth, we should just have (total Earth flux) × S(θ), so the angular profile is just a property of the instrument and the amplitude falls as the Earth's solid angle. HST is different as in that case, the source is resolved on the scale of the scattering function.

lpsinger: Now that I understand the physics better... we could probably do something more general if we took the solar spectrum x simplified albedo model x Earth spheroid x low-order integrator x a form factor for the off-axis scattering function.

We should see if there is some prior art for doing this calculation.

mcoughlin: I think the problem is getting the albedo wrong is a really big deal, so you would need real UV radiative transfer, and instead the empirical STIS spectrum already encodes that for us. So to me I like the idea of doing extended-source integration over the illuminated spheroid with a Point Source Transmittance. The ULTRASAT paper says they have this measured with the baffle achieving better than 2×10⁻¹¹ "stray light suppression", defined in their footnote (1) as:

the ratio between the scattered photon flux produced by a source (e.g., Earth) on the detector and the source photon flux at geostationary orbit

So we could use this as the expected suppression.

lpsinger: Sure, so you're just lumping the (solar spectrum x simplified albedo model) into the STIS spectrum, and then you have specular reflection from the surface of the Earth. Right?

Worth a shot.

lpsinger: @hpearnshaw what do you think?

mcoughlin: @lpsinger With that technique, I get 0.0362 e⁻/pix against their 12, so a factor of 331... which is better than 10^9 ;)

lpsinger: I stuck a print statement in my little gallery script and I see that the scale factor attains maximum values of about 500. Are you sure that the normalization is working?

src/m4opt/synphot/background/_earthshine/__init__.py
```diff
@@ -0,0 +1,383 @@
...
+# See https://etc.stsci.edu/etcstatic/users_guide/1_ref_9_background.html
+_LIMB_ANGLES_DEG = np.array([24.0, 38.0, 50.0])
+_LOG2_SCALE_FACTORS = np.array([1.0, 0.0, -1.0])  # log2([2.0, 1.0, 0.5])
+
+# Altitude at which those calibration points were measured. The angles above
+# are meaningful only relative to how large the Earth looked from there.
+_HST_ALTITUDE = 540 * u.km
+_HST_ANGULAR_RADIUS_DEG = np.degrees(
+    np.arcsin((R_earth / (R_earth + _HST_ALTITUDE)).to_value(u.dimensionless_unscaled))
+)
```
lpsinger:
```suggestion
_HST_ANGULAR_RADIUS_DEG = np.arcsin(R_earth / (R_earth + _HST_ALTITUDE)).to_value(u.deg)
```

src/m4opt/synphot/background/_earthshine/__init__.py
```diff
@@ -0,0 +1,383 @@
...
+
+def _earth_angular_radius_deg(observer_location, obstime):
+    """Angular radius of the Earth as the observer sees it, in degrees."""
+    radius = observer_location.get_gcrs(obstime).cartesian.norm()
+    with np.errstate(invalid="ignore"):
+        return np.degrees(
+            np.arcsin(
+                np.clip((R_earth / radius).to_value(u.dimensionless_unscaled), 0, 1)
+            )
+        )
```
lpsinger: This function is unused.

```suggestion
```

src/m4opt/synphot/background/_earthshine/__init__.py
```diff
@@ -0,0 +1,383 @@
...
+from ....constraints._earth_limb import _get_angle_from_earth_limb
+from ..._extrinsic import ExtrinsicScaleFactor
+from .._core import BACKGROUND_SOLID_ANGLE
+from . import data
+
+# Calibration points from HST STIS Instrument Handbook, Table 6.4, and
+# the STScI ETC documentation.
+# See https://etc.stsci.edu/etcstatic/users_guide/1_ref_9_background.html
+_LIMB_ANGLES_DEG = np.array([24.0, 38.0, 50.0])
+_LOG2_SCALE_FACTORS = np.array([1.0, 0.0, -1.0])  # log2([2.0, 1.0, 0.5])
```
lpsinger: This is unused.

```suggestion
```

lpsinger (changes requested):

src/m4opt/synphot/background/_earthshine/__init__.py
```diff
@@ -0,0 +1,383 @@
...
+    poor substitute for a ray trace of a particular baffle design, and the further an
+    observatory is from the conditions under which the STIS numbers were measured
+    -- an observatory at geostationary orbit, for instance -- the less the
+    absolute normalization should be trusted. Use the ``factor`` argument to
+    renormalize the model to an observatory's own stray light budget.
+
+    Three separate things limit how far from the Earth this model may be used.
+    The geometry of the limb itself is exact at any distance, but:
+
+    - The point source transmittance is one power law, with an exponent fixed
```
lpsinger, line 446:
```suggestion
    - The oblateness of the Earth is neglected.

    - The point source transmittance is one power law, with an exponent fixed
```

lpsinger (changes requested):

src/m4opt/missions/_ultrasat/__init__.py
mcoughlin: The normalization is fine, but the point source transmittance diverges on axis and my 1° floor was arbitrary, so instead I think it is better to use 24°, the smallest angle the STIS points constrain, which drops your gallery maximum to 1.30.

lpsinger: Does that affect your matching factor at all? Does it bring it closer to unity?

mcoughlin: The factor is the same as the floor only touches lines of sight within 24° of the limb, and ULTRASAT's constraint is 48°.

Since the factor is the ratio of two instruments' off-axis rejection, I don't think there is a reason for two different telescopes to agree. Unity would mean ULTRASAT's baffle scatters exactly like HST's.

However, the idea that ULTRASAT scatters 331× more than HST seems .... unlikely. More likely, I would guess that we're not modelling the same geometry they are. Their worst case is "full Earth at the closest allowed angle from the lowest side of the baffle", whereas we're measuring from the optical axis. I am sending an email to Yossi (ccing you) asking them which angle their 12 e⁻ is referenced to.

lpsinger:
Would you please merge https://github.com/mcoughlin/m4opt/pull/1 into your branch?

src/m4opt/synphot/background/_earthshine/__init__.py
```diff
@@ -0,0 +1,376 @@
...
+    Examples
+    --------
+
+    Constant "high" earthshine spectrum (no spatial dependence):
+
+    >>> from astropy import units as u
+    >>> from m4opt.synphot.background import EarthshineBackground
+    >>> background = EarthshineBackground.high()
+    >>> float(background(5000 * u.angstrom).value) > 0
+    True
```
lpsinger: This is not a very satisfying example!

```suggestion
    >>> float(background(5000 * u.angstrom).value)
    ...put value here...
```

lpsinger (changes requested):

src/m4opt/synphot/background/_earthshine/__init__.py
mcoughlin: oops fixed

src/m4opt/missions/_ultrasat/__init__.py
lpsinger: Would you please make an issue to follow up on that? Otherwise this looks great.

lpsinger:
Readthedocs build is timing out. Any low-hanging fruit for speeding up the calculation?

mcoughlin:
@lpsinger sorry pushed a fix.

lpsinger (approved):

PR #441, opened 2026-02-21
Add Gurobi as backend option

mcoughlin:
This PR adds Gurobi as a supported backend in M4OPT.

Unfortunately, the Academic CPLEX version we can download only has Python bindings up until 3.10, so is a problem for any university folks who want to do runs on recent M4OPT versions.


lpsinger (commented):
I like the idea of having a vendor-neutral MILP interface that can plug in to Gurobi or CPLEX, but I have some concerns about this approach.

The `milp.Model` class is a light wrapper around `docplex.mp.model.Model` that fixes some bugs in it and addresses some obvious shortcomings such as Numpy broadcasting. Now you have written a wrapper for Gurobi that gives it a CPLEX-like interface, but why do we want Gurobi to look like CPLEX?

It would be better to design a Numpy-friendly, Python-friendly wrapper for both solvers without being a slave to the quirks of the docplex API.

Better still would be to identify an existing third-party project that we could reuse. I think that there are several good multi-vendor MILP libraries out there. Did you review the current state of the art?

The Gurobi installation instructions in the M4OPT documentation have a link to an IBM page that tells you how to install CPLEX in newer versions of Python. Do those instructions work? If not, then how do they fail?

mcoughlin:
@lpsinger It is unclear to me that academic users have access to that link. I spent quite a while trying to get to it and gave up. If you can, maybe we can report the upshot of it in the docs?

lpsinger:
That was a remarkably fast and detailed reply. Was it generated using AI?

By what criterion did you assess whether a framework was Numpy friendly?

mcoughlin:
@lpsinger I had gone through this exact question this morning, and formatting and discussion is helped by chatgpt. The Numpy assessment is basically if model.add_constraints_(A @ x >= b) where A is a NumPy array. I went through this with gwemopt and used pulp: https://github.com/coin-or/pulp (https://github.com/skyportal/gwemopt/blob/main/gwemopt/utils/milp.py) but strugged a lot with it.

mcoughlin:
@lpsinger i am trying to say that i cannot get access to those instructions either with my account level, which at least makes it unhelpful for any of my people. 

I emailed their help and they said something about company access blah blah, so I really don't see that as a pathway forward.

mcoughlin:
@lpsinger It's got my EDU and everything, but I might misunderstand how their account stuff works (but it was honestly making me very unhappy so I rage quit at some point).

PR #445, opened 2026-02-23
Update CPLEX academic installation instructions

lpsinger:


mcoughlin (approved):
Tested and works on both Mac and Linux.

PR #448, opened 2026-03-01
Ensure appmag_dist also checks for adaptive_exptime

mcoughlin:
This PR adds a check for adaptive_exptime when using appmag_dist.


lpsinger (changes requested):
Would you please add a unit test that exercises the configuration that caused this to fail?

mcoughlin:
@lpsinger Done

lpsinger (approved):

PR #450, opened 2026-03-01
Warm start solution with non-zero fields

mcoughlin:
This PR improves solver efficiency by warm starting the MILP solver with fields that have non-zero probability.

### DIFF
--- a/m4opt/_cli/schedule.py
+++ b/m4opt/_cli/schedule.py
@@ -68,6 +68,20 @@ def invert_footprints_to_regions(footprints, n_pixels):
     return pixel_to_region_map, region_to_fields_map
 
 
+def prefilter_fields(hpx, skymap_flat, target_coords, fov_radius):
+    """Return boolean mask of fields overlapping nonzero sky map probability."""
+    from astropy.coordinates import search_around_sky
+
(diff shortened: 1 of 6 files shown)


lpsinger (commented):
This isn't a "warm start"; that would be [initializing the solver with a partial solution](https://www.ibm.com/docs/en/icos/22.1.0?topic=mip-starting-from-solution-starts), something that is supported by CPLEX but that we are not using at the moment.

How does this improve solver efficiency? As far as I can tell it will result in the same problem setup because there are later filtering steps.

I can see that this might speed up the problem setup, but by how much?

Do our sky maps actually contain zero values? As far as I can recall I think that BAYESTAR sky maps do not. Maybe Fermi GBM sky maps do?

m4opt/_cli/schedule.py
```diff
@@ -206,6 +237,14 @@ def schedule(
             rich_help_panel="Solver Options",
         ),
     ] = None,
+    prefilter: Annotated[
```
lpsinger:
```suggestion
    max_credible_level: Annotated[
```

lpsinger (changes requested):

PR #457, opened 2026-03-11
Don't let codespell automatically fix up code

lpsinger:
Automatic spell check is much more likely to accidentally make unwanted functional changes to code, especially when short variable names are misinterpreted as missspelled English words. It's safer to display the spell check errors and let the contributor deal with them manually.


mcoughlin (approved):
LGTM

PR #459, opened 2026-03-11
[WIP] Add gurobi

mcoughlin:
Gurobi addition on my branch, including Leo's test from https://github.com/m4opt/m4opt/pull/458


lpsinger (commented):
I remain reluctant to maintain our own multi-solver MILP front end, although I can be convinced of this approach.

I would rather reuse and contribute to an existing solver front-end, and I recall you found one that you liked and opened a few PRs.

If we did write our own multi-solver compatibility layer, then I would rather _not_ make other solvers look like docplex. docplex has a terrible API.

mcoughlin:
@lpsinger my gut says for now we should make our own and maybe eventually move elsewhere. Is there an API you don't dislike that you know of?

PR #461, opened 2026-03-19
Add support for HEALPix footprints of non-convex polygons

lpsinger:
This is going to be useful for generating HEALPix footprints of the UVEX survey footprint regions.


mcoughlin (approved):
LGTM

PR #473, opened 2026-04-03
Add support for partitioning graphs with node and edge weights

lpsinger:
Also, drop support for graph input as an adjacency matrix beause it would be complicated to support appropriate arguments for node and edge weights.


mcoughlin (approved):
LGTM

PR #478, opened 2026-04-26
Add method to orient UVEX for a ground pass

lpsinger:


m4opt/missions/_uvex.py
```diff
@@ -162,3 +163,194 @@
...
+    -------
+    :
+        Target coordinate and roll angle.
+
+    Notes
+    -----
+    UVEX's high gain antenna is at a 45° angle to the spacecraft axes, and
+    points toward the cool side and opposite the telescope boresight. For a
+    ground contact, the following conditions should be met, as shown in the
+    figur below:
```
mcoughlin: figure

m4opt/missions/_uvex.py
```diff
@@ -162,3 +163,194 @@
...
+        ax.legend(title="Separation", loc="center right", bbox_to_anchor=(0.975, 0.25))
+    """
+    observer_location = uvex.observer_location(time)
+    sun_coord = get_body("sun", time, observer_location)
+    earth_coord = get_body("earth", time, observer_location)
+    anti_earth_coord = SkyCoord(-earth_coord.cartesian, frame=earth_coord.frame)
+    roll = nominal_roll(observer_location, anti_earth_coord, time)
+    offset_frame = anti_earth_coord.skyoffset_frame(roll)
+    target_coord = SkyCoord(0 * u.deg, -45 * u.deg, frame=offset_frame)
+    oops = target_coord.separation(sun_coord) <= 45 * u.deg
```
mcoughlin: Maybe flip_needed or similar would be a good name?

m4opt/missions/_uvex.py
```diff
@@ -162,3 +163,194 @@
...
+    target_coord = SkyCoord(
+        0 * u.deg, np.where(oops, 45, -45) * u.deg, frame=offset_frame
+    ).transform_to(earth_coord.frame)
+    roll = nominal_roll(observer_location, target_coord, time)
```
mcoughlin, line 366: Does roll always need to be computed twice?

m4opt/missions/_uvex.py
```diff
@@ -162,3 +163,194 @@
...
+        ax.plot(dt, antenna.separation(earth), label=r"Target $\leftrightarrow$ antenna")
+        ax.legend(title="Separation", loc="center right", bbox_to_anchor=(0.975, 0.25))
+    """
+    observer_location = uvex.observer_location(time)
+    sun_coord = get_body("sun", time, observer_location)
+    earth_coord = get_body("earth", time, observer_location)
+    anti_earth_coord = SkyCoord(-earth_coord.cartesian, frame=earth_coord.frame)
+    roll = nominal_roll(observer_location, anti_earth_coord, time)
+    offset_frame = anti_earth_coord.skyoffset_frame(roll)
+    target_coord = SkyCoord(0 * u.deg, -45 * u.deg, frame=offset_frame)
```
mcoughlin: Maybe add a comment here on what is being checked exactly.

mcoughlin (approved):
A few things on variable names and comments.

m4opt/missions/_uvex.py
lpsinger: Fixed.

m4opt/missions/_uvex.py
lpsinger: Fixed, although I admit I worked that part out by trial and error.

m4opt/missions/_uvex.py
lpsinger: It does. We are first calculating a roll angle that orients the telescope toward the Earth, and then pitching to point the antenna toward the Earth. Picture the solar panels remaining stationary and the spacecraft itself pitching within them.

After this rotation the telescope boresight is pointing in a new direction, so the roll angle about that new direction is different.

There is probably a simple formula to calculate the new roll angle, but it didn't come to me. This is the lazy, brute force solution. I am not proud of it. Do you see a simpler solution?

m4opt/missions/_uvex.py
lpsinger: It's ensuring that the sun is at least 45° away from the telescope boresight, which should now be explained by the names of the variables.

m4opt/missions/_uvex.py
mcoughlin: Yeah I figured. I didn't see immediately what to do either

lpsinger:
Heads up that I just added a little bit more documentation.

mcoughlin (approved):
LGTM

PR #492, opened 2026-06-11
Add ground-based telescope model to _slew.py

devas015:


m4opt/dynamics/_slew.py
```diff
@@ -218,3 +218,228 @@ def separation(
...
+
+        Returns
+        -------
+        min_dist:
+            Shortest possible angular separation of the two positions.
+        """
+        if not init_pos.unit.is_equivalent(fin_pos.unit):
+            raise ValueError(
+                "Initial and final position values should be in the same angular units."
+            )
```
lpsinger: Remove unnecessary validation; unit conversion will fail if the units are not compatible.

m4opt/dynamics/_slew.py
```diff
@@ -218,3 +218,228 @@ def separation(
...
+    """Maximum angular jerk. Defaults to None if a value isn't specified."""
+
+    settling_time: u.Quantity[u.physical.time] = 0 * u.s
+    """Time to settle to rest after a slew."""
+
+    @staticmethod
+    def separation(
+        init_pos: u.Quantity[u.physical.angle],
+        fin_pos: u.Quantity[u.physical.angle],
+        motion_range: tuple[float, float] | None = None,
```
lpsinger: This should be an instance data member, not an argument. Also, it should have units.

m4opt/dynamics/_slew.py
```diff
@@ -218,3 +218,228 @@ def separation(
...
+
+    @staticmethod
+    def _movetime(
+        dist: u.Quantity[u.physical.angle],
+        vmax: u.Quantity[u.physical.angular_velocity],
+        amax: u.Quantity[u.physical.angular_acceleration],
+        jmax: u.Quantity | None = None,
+        settling_time: u.Quantity[u.physical.time] = 0 * u.s,
+    ) -> u.Quantity[u.physical.time]:
+        """See https://www.jpe-innovations.com/precision-point/third-order-point-to-point-motion-profile/"""
```
lpsinger: Is it necessary to implement maximum jerk in order to model Rubin's slew time?

If so, then please refactor so that both the eigenaxis and ground-based slew models have this feature without duplicated code.

Have you checked the math on that blog post?

m4opt/dynamics/_slew.py
```diff
@@ -218,3 +218,228 @@ def separation(
...
+
+        Returns
+        -------
+        :
+            Time to slew between the initial and final positions.
+        """
+
+
+@dataclass
+class GroundSlewComponent(GroundSlew):
```
lpsinger: I don't see why you need an abstract base class.

```suggestion
@dataclass
class GroundSlewComponent:
```

lpsinger (changes requested):

m4opt/dynamics/_slew.py
devas015: I have verified the math from the blog post, yes. Rubin's slew model implements both infinite and finite jerk cases. Infinite jerk underestimates actual slew time by up to 0.3s for Rubin's kinematic parameters. I wasn't sure whether or not the 0.3s is considered negligible. If it is negligible, I can remove the finite jerk case.

lpsinger:
This has merge conflicts. Would you please rebase it?

lpsinger:
Still has merge conflicts.

m4opt/dynamics/_slew.py
```diff
@@ -302,3 +303,145 @@ def separation(
             @ rotation_matrix(center2.lat, "y")
         )
         return Angle(np.arccos(0.5 * (matrix_trace(mat) - 1)) * u.rad).to(u.deg)
+
+
+@dataclass
+class GroundComponentSlew(AngularMotionProfile):
```
lpsinger:
```suggestion
class SlewComponent(AngularMotionProfile):
```

m4opt/dynamics/_slew.py
```diff
@@ -302,3 +303,145 @@ def separation(
...
+            raise ValueError(
+                "If the telescope is AltAz, a location and observation time must be specified."
+            )
+        if self.coord_sys_is_equatorial:
+            coord1 = coord1.transform_to("icrs")
+            coord2 = coord2.transform_to("icrs")
+            time1 = self.comp1.time(coord1.ra, coord2.ra)
+            time2 = self.comp2.time(coord1.dec, coord2.dec)
+            time3 = self.comp3.time(coord1.ra, coord2.ra) if self.comp3 else (0 * u.s)
+            time4 = self.comp4.time(coord1.dec, coord2.dec) if self.comp4 else (0 * u.s)
```
lpsinger: Although the telescope might have an equatorial mount, the dome almost always horizontal (i.e., works in the alt-az frame).

m4opt/dynamics/_slew.py
```diff
@@ -302,3 +303,145 @@ def separation(
...
+        else:
+            coord1 = coord1.transform_to(
+                AltAz(obstime=time_obs, location=self.location)
+            )
+            coord2 = coord2.transform_to(
+                AltAz(obstime=time_obs, location=self.location)
+            )
```
lpsinger: Define the frame only once:

```suggestion
            frame = AltAz(obstime=time_obs, location=self.location)
            coord1 = coord1.transform_to(frame)
            coord2 = coord2.transform_to(frame)
```

m4opt/dynamics/_slew.py
```diff
@@ -302,3 +303,145 @@ def separation(
...
+    comp4: GroundComponentSlew | None = None
+    """Components of the telescope separated by part and axis (e.g. mount 
+    in altitude, mount in azimuth, dome in altitude, and dome in azimuth 
+    may make up the four components). Only two are required. Odd 
+    components must be along the RA or Alt axes while even components 
+    must be along the Dec or Az axes."""
+
+    coord_sys_is_equatorial: bool = True
+    """If the telescope works in equatorial or altaz coordinates. Defaults
+    to equatorial coordinates. Units do not have to match the skymap's."""
```
lpsinger: Instead of a flag, please define two derived classes, `EquatorialSlew` and `AltAzSlew`. They can each have a method called `_frame` which returns the frame of the telescope mount.

m4opt/dynamics/_slew.py
```diff
@@ -302,3 +303,145 @@ def separation(
...
+        """
+        if not self.coord_sys_is_equatorial and (
+            (self.location is None) or (time_obs is None)
+        ):
+            raise ValueError(
+                "If the telescope is AltAz, a location and observation time must be specified."
+            )
+        if self.coord_sys_is_equatorial:
+            coord1 = coord1.transform_to("icrs")
+            coord2 = coord2.transform_to("icrs")
```
lpsinger: Actually this should be converted to a GCRS frame that is at the time of the observation and the location of the telescope.

lpsinger (changes requested):

devas015:
I added 3 derived classes to account for telescopes that are fully in AltAz, fully equatorial, or have a mix of the two (e.g. if the mount was equatorial but the dome was AltAz).

lpsinger:
This branch shows 145 files changed. I am guessing that it didn't rebase cleanly after #537. Would you please clean it up?

src/m4opt/missions/_rubin/__init__.py
```diff
@@ -42,6 +42,34 @@ def _make_fov():
     )
 
 
+# Initialize Components for Rubin's Slew Model
```
lpsinger, line 45: Add comments to indicate the provenance of these numbers.

lpsinger (changes requested):
Please change the generic names `comp1`, `comp2`, etc. to descriptive names that identify the axes.

I'm not a fan of the `if` statements on coordinate index. What if the slew component knows the frame and coordinate component it's supposed to use?

lpsinger:
Hi @devas015, any updates on this?

devas015:
When you say "what if the slew component knows [its] coordinate component," do you mean adding the coordinate component as an attribute to `SlewComponent`? I can't find another way to access the alt/az/ra/dec components of each coordinate without `.alt`, `.az`, etc, so my concern is that the `if` statements would just get moved to `SlewComponent` (albeit as shorter `if` statements), unless you know of another way to implement this?

lpsinger:
> When you say "what if the slew component knows [its] coordinate component," do you mean adding the coordinate component as an attribute to `SlewComponent`? I can't find another way to access the alt/az/ra/dec components of each coordinate without `.alt`, `.az`, etc, so my concern is that the `if` statements would just get moved to `SlewComponent` (albeit as shorter `if` statements), unless you know of another way to implement this?

Using `getattr`.

devas015:
Components can now be any axis or coordinate system, and sources have been added for the Rubin numbers.

src/m4opt/dynamics/_slew.py
```diff
@@ -302,3 +309,257 @@ def separation(
...
+    telescope is AltAz, a location must be specified.
+    """
+
+    location: EarthLocation
+    """Location of the observatory."""
+
+    comp1: SlewComponent
+    comp2: SlewComponent
+    comp3: SlewComponent | None = None
+    comp4: SlewComponent | None = None
```
lpsinger: Just make this a list of components.

PR #498, opened 2026-06-15
Rename BangBangTrajectory to AngularMotionProfile

lpsinger:
This is not really a bang-bang trajectory; that refers to the optimal control strategy for a system with an actuator that only has two states, on or off.

Move the documentation and plots to the AngularMotionProfile docstring.

### DIFF
--- a/m4opt/dynamics/__init__.py
+++ b/m4opt/dynamics/__init__.py
@@ -7,6 +7,6 @@ this package is more organized.
 """
 
 from ._roll import nominal_roll
-from ._slew import EigenAxisSlew, Slew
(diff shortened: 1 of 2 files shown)


devas015:
This looks good to me!

lpsinger:
@devas015, what is the URL of the web page from which you got the equations? I want to include it in the documentation.

devas015:
@lpsinger I got it from here: https://www.jpe-innovations.com/precision-point/third-order-point-to-point-motion-profile/

It is also posted here: https://www.researchgate.net/publication/289374755_THIRD_ORDER_POINT-TO-POINT_MOTION_-PROFILE

PR #503, opened 2026-06-25
Add constraints to enforce visiting all fields before revisiting

devas015:
This enforces every field being visited k - 1 times before any field being visited k times (hopefully). This assumes scalar exposure times and delays as described by Equations 3 and 4 [here](https://www.overleaf.com/read/jrmmrbvfxzft#e17595). 

Once this is edited and merged, I can open PRs to account for sequences of exposure times and delays as described by Equations 5, 6, and 7 in the linked document.

### DIFF
--- a/docs/design/milp.rst
+++ b/docs/design/milp.rst
@@ -95,10 +95,10 @@ Constraints
     :label: fixed-exptime-constraint-no-overlap
 
     \begin{eqnarray}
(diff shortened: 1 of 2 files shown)


lpsinger (changes requested):
We need to retain the existing problem formulation which is more appropriate for observatories lacking a filter changer, and _add to_ rather than _replace_ it. So the design document will need a new section that describes the additional scenario, how it is different than the existing scenarios, and how the equations are altered. Likewise the scheduler code will need a command line option to control this and will need to have a matching `if` statement somewhere.

However, it would be good to try out what you have on this branch on a few example sky maps first, and make sure that they come out how you think they should!

Issue #514, opened 2026-07-06
Schedules written by `m4opt schedule` cannot be read back (`observer_location` column not deserializable)


wathela:
For Earth-fixed missions, `m4opt schedule` writes an `observer_location`
column of class `EarthFixedObserverLocation`. Reading the ECSV back with
astropy — including by `m4opt animate` itself — fails with:

ValueError: unsupported class for construct m4opt.observer._earth_fixed.EarthFixedObserverLocation

because the class is not in astropy's hard-coded serialization whitelist
(`astropy.table.serialize.__construct_mixin_classes`), so schedules for
ground-based missions cannot be round-tripped at all.

**To reproduce:**

```python
from astropy.table import QTable
QTable.read("schedule.ecsv")   # schedule produced with --mission ztf
```

**Workaround:** rewriting the header's `__class__` from
`m4opt.observer._earth_fixed.EarthFixedObserverLocation` to
`astropy.coordinates.earth.EarthLocation` makes the file readable — the
serialized attributes (ellipsoid, x, y, z) are identical since the class
subclasses `EarthLocation`.

**Possible fix:** downcast the column to plain `EarthLocation` in the
schedule writer, or register the class with astropy's serialization
machinery.

Environment: Python 3.12, astropy 8.0.0, current main.

PR #516, opened 2026-07-06
Fix `m4opt animate` for Earth-fixed (ground-based) missions

### LINKED ISSUE #515: `m4opt animate` fails for Earth-fixed (ground-based) missions
`m4opt animate` crashes on schedules produced for ground-based missions
(e.g. `--mission ztf`), in two places:

1. `EarthFixedObserverLocation.__call__` returns the scalar site regardless
   of the time argument, but `animate` indexes the result per timestep
   (`observer_locations[0]` at animate.py:211, and
   `observer_locations[:, np.newaxis]` at animate.py:247):

TypeError: 'EarthFixedObserverLocation' object with a scalar value does not support indexing

2. After fixing (1), the footprint patch construction assumes each footprint
   region has `.vertices` (animate.py:276), but ZTF's FOV is a `Regions`
   collection (the 64-quadrant CCD mosaic):

AttributeError: 'Regions' object has no attribute 'vertices'

**To reproduce:** schedule any event with `--mission ztf` (fixed exposure
mode is enough), then run `m4opt animate` on the output. Note that
reproduction currently also requires working around #514 in
order to read the schedule file at all.

With both fixed, animation renders correctly end-to-end for ZTF
(field of regard, observation footprints, frame rendering). PR incoming.

Environment: Python 3.12, astropy 8.0.0, current main.

wathela:
Fixes #515.

Two changes:

1. **`m4opt/observer/_earth_fixed.py`** — `EarthFixedObserverLocation.__call__`
   broadcasts the fixed site to the time argument's shape via
   `np.broadcast_to(self, time.shape, subok=True)` (a zero-copy view),
   matching `TleObserverLocation`'s contract that `observer_location(time)`
   returns locations with the same shape as `time`. Scalar `Time` behavior
   and the existing doctest are unchanged.

2. **`m4opt/_cli/animate.py`** — the footprint patch construction now
   handles FOVs that are `Regions` collections (e.g. ZTF's 64-quadrant CCD
   mosaic) by iterating sub-regions, in both the main map and the zoom
   inset. The per-observation nesting is preserved so the frame-by-frame
   visibility toggling still pairs one patch group per schedule row.

**Testing:** verified end-to-end on a real Fermi-GBM sky map scheduled with
`--mission ztf` (fixed exposure mode). All animate stages complete —
field of regard, observation footprints, frame rendering — and the GIF
renders correctly. Also re-verified that
`m4opt schedule` still runs unchanged for ZTF after the observer-location
change.

Note: reproducing the bug (and testing this fix) currently requires working
around #514 to read the schedule file back.


mcoughlin:
@wathela please fix the tests.

wathela:
@mcoughlin  The CI failure is the `TleObserverLocation.from_id` doctest timing out on a
network call to celestrak.org — unrelated to this PR (no changes touch
`_tle.py`). Should clear on re-run.

Also pushed a fix for the zoom inset: the footprint patches were built but
never added to the inset axes, so it rendered empty. Verified end-to-end with
`--mission ztf`, footprints now render and toggle in both the main map and
the inset.
test_anim_inset

wathela:
@mcoughlin CI is failing on the `TleObserverLocation.from_id` doctest again — it fetches
a TLE from celestrak.org at test time and the runner is timing out on that
connection (same failure on both runs). Nothing in this PR touches `_tle.py`;
everything I changed passes, with `_earth_fixed.py` at 100% coverage.

Would you like me to mark that doctest as remote-data in a separate PR so
celestrak outages stop failing CI? Happy to also re-trigger this run once
celestrak is reachable.

mcoughlin:
A separate PR for that seems fine @wathela

devas015 (commented):
This looks good to me! I tested it with a couple Rubin observing plans as well and it seemed to work fine.

lpsinger (approved):

PR #525, opened 2026-07-29
Add helioecliptic longitude constraint

lpsinger:
UVEX is evaluating a possible survey policy of scheduling survey fields at times near lowest zodiacal light conditions. Zodiacal light is primarily a function of ecliptic latitude and of the difference in ecliptic longitude between the target and the sun. For fixed stars, the former does not change significantly over human timescales, whereas the latter varies through a full cycle over the course of the year.

### DIFF
--- a/m4opt/constraints/_positional.py
+++ b/m4opt/constraints/_positional.py
@@ -3,15 +3,18 @@
 from abc import abstractmethod
 from typing import override
 
+import numpy as np
 from astropy import units as u
 from astropy.coordinates import (
     ICRS,
     AltAz,
     Angle,
     EarthLocation,
+    GeocentricTrueEcliptic,
     HADec,
     SkyCoord,
     UnitSphericalRepresentation,
+    get_sun,
 )
 from astropy.time import Time
 
@@ -63,6 +66,14 @@ class HADecConstraint(AngleConstraint):
     @override
     def _frame(self, observer_location, obstime):
         return HADec(obstime=obstime, location=observer_location)
+
+
+class GeocentricTrueEclipticConstraint(AngleConstraint):
+    """Constrain an angle in the :class:`~astropy.coordinates.GeocentricTrueEclipticConstraint` frame."""
+
+    @override
+    def _frame(self, observer_location, obstime):
+        return GeocentricTrueEcliptic(obstime=obstime)
 
 
 class ICRSConstraint(AngleConstraint):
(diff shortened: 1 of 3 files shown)


m4opt/constraints/_positional.py
```diff
@@ -158,3 +169,22 @@ class HourAngleConstraint(LongitudeConstraint, HADecConstraint):
     --------
     RightAscensionConstraint
     """
+
+
+class HelioeclipticLongitudeConstraint(GeocentricTrueEclipticConstraint):
+    """Constraint the helioecliptic longitude of the target.
+
+    This places a constraint on the absolute value of the ecliptic longitude of
```
mcoughlin: Note the allowed range?

m4opt/constraints/_positional.py
```diff
@@ -158,3 +169,22 @@ class HourAngleConstraint(LongitudeConstraint, HADecConstraint):
...
+    the target minus the ecliptic longitude of the sun.
```
mcoughlin: Since it is an absolute magnitude, it is an elongation, so might rename?

m4opt/constraints/_positional.py
```diff
@@ -158,3 +169,22 @@ class HourAngleConstraint(LongitudeConstraint, HADecConstraint):
...
+    """
+
+    def _get_angle(self, observer_location, target_coord, obstime):
```
mcoughlin, line 189: Missing override.

mcoughlin (changes requested):
Could use a test.

m4opt/constraints/_positional.py
lpsinger: Fixed.

m4opt/constraints/_positional.py
lpsinger: Fixed.

m4opt/constraints/_positional.py
lpsinger: This is a new term to me. Based on https://en.wikipedia.org/wiki/Elongation_(astronomy), it seems that elongation is only something that planets or other solar system bodies have. What do you think it means?

mcoughlin: I was thinking of it as basically the term for angular separation from the Sun, which I realize this is almost but not quite the same as :)

lpsinger:
I added tests. Please review.

mcoughlin (approved):
LGTM

PR #527, opened 2026-07-31
EarthFixedObserverLocation shouldn't subclass of EarthLocation

### LINKED ISSUE #514: Schedules written by `m4opt schedule` cannot be read back (`observer_location` column not deserializable)

lpsinger:
A "has-a" relationship is more appropriate than an "is-a" relationship here.

Fixes #514. Fixes #526.

### DIFF
--- a/m4opt/missions/_rubin/__init__.py
+++ b/m4opt/missions/_rubin/__init__.py
@@ -3,7 +3,7 @@ from importlib import resources
 import numpy as np
 import yaml
 from astropy import units as u
-from astropy.coordinates import SkyCoord
+from astropy.coordinates import EarthLocation, SkyCoord
 from astropy.table import Table
 from regions import RectangleSkyRegion, Regions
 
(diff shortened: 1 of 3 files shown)


devas015 (commented):
This looks good to me.

PR #530, opened 2026-07-31
Skip doctest that contacts Celestrak

lpsinger:
Celestrak has severe rate limits and a draconian policy around enforcing them. The tests that depend on Celestrak are flaky. Disable them for now.

This is not a great loss right now, because due to the recent exhaustion of 5-digit catalog numbers, we need to migrate from TLEs to a new format (see https://celestrak.org/NORAD/documentation/gp-data-formats.php).


mcoughlin (approved):
LGTM

PR #533, opened 2026-07-31
Update cached directories for Astropy 8.0

lpsinger:
See https://docs.astropy.org/en/stable/whatsnew/8.0.html#overhaul-of-cache-and-configuration-path-discovery-mechanism


mcoughlin:
@lpsinger weird CI failure.

lpsinger:
> @lpsinger weird CI failure.

Yes, see #534.

PR #523, opened 2026-07-28
Add UVEX chip gaps

lpsinger:
Data provided by @hpearnshaw.

PR #536, opened 2026-07-31
Adjust UVEX sky grid for new FOV model

lpsinger:
This past Tuesday I got a model of the imaging FOV that includes chip gaps (see #523). The bounding rectangle is about 3.66° x 3.74°, slightly larger than the 3.5° square that I had been using before. We can decrease the number of tiles in the sky grid accordingly.


mcoughlin (approved):
LGTM

PR #539, opened 2026-08-04
Add function to calculate size of 1D intersection

lpsinger:
This is needed to calculate the cadence distribution of a survey while accounting for overlap between fields.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,8 +17,7 @@ on:
 
 jobs:
 
-  build:
-    name: Build
+  sdist:
     permissions: {}
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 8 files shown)


lpsinger:
Blocked by https://github.com/astropy/extension-helpers/pull/155.

src/m4opt/utils/numpy.py
```diff
@@ -1,6 +1,11 @@
 """Utilities for working with Numpy arrays."""
 
 import numpy as np
+from numpy import typing as npt
+
+from ._numpy import count_intersect1d as _count_intersect1d
+
+__all__ = ("atmost_1d", "clump_nonzero", "clump_nonzero_inclusive", "count_intersect1d")
```
mcoughlin: Is full_indices gone?

.github/workflows/publish.yml
```diff
@@ -17,32 +17,50 @@ on:
...
 jobs:
 
-  build:
-    name: Build
+  sdist:
     permissions: {}
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
```
mcoughlin, line 26: I assume these are stale? Or is the downgrade purposeful?

src/m4opt/utils/_numpy.c
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+        p++;
+    }
+}
+
+static PyObject *py_count_intersect1d(PyObject *NPY_UNUSED(self), PyObject *const *args, Py_ssize_t nargs) {
+    PyObject *a = NULL, *b = NULL, *result = NULL;
+    if (nargs != 2) {
+        PyErr_Format(PyExc_TypeError, "count_intersect1d() takes exactly one argument (%zd given)", (ssize_t) nargs);
```
mcoughlin: two arguments

mcoughlin (changes requested):
A few comments

src/m4opt/utils/numpy.py
lpsinger: No, that was my mistake.

.github/workflows/publish.yml
lpsinger: No, it was a copy-paste from another project.

src/m4opt/utils/_numpy.c
lpsinger: Fixed.

mcoughlin (approved):
LGTM

PR #546, opened 2026-08-15
Add sigfigs to chip gap dimensions

lpsinger:
This is according to a diagram received from Hannah Earnshaw on 2026-08-18.

### DIFF
--- a/src/m4opt/missions/_uvex/__init__.py
+++ b/src/m4opt/missions/_uvex/__init__.py
@@ -24,7 +24,7 @@ _chip_length = 40.96 * u.mm
 _chip_angle = 4219 * u.arcsec
 """Linear extent of one imager chip in angular units"""
 
-_chip_gap_length = np.asarray([2.5, 4]) * u.mm
+_chip_gap_length = np.asarray([4.05, 2.55]) * u.mm
 """Linear extents of chip gaps in length units"""
 
 _chip_offset = [-1, 0, 1] * (


bwgref (commented):
Please fix commit message for proper attributions. There's no way this came to Sterl via Elias Most becuase neither one of those people are working on the technical aspects of the Payload (or UVEX at all in the case of Elias Most).

lpsinger:
> Please fix commit message for proper attributions. There's no way this came to Sterl via Elias Most becuase neither one of those people are working on the technical aspects of the Payload (or UVEX at all in the case of Elias Most).

OK, how should I attribute it?

Sterl did email this to us; I don't know the provenance before that.

lpsinger:
Ah, Elias *Kammoun*. Elias *Most* was my office mate the last time I went to Aspen.

Is this correct now? And are the numbers correct?

hpearnshaw (changes requested):
X is the larger gap size and Y the smaller one in this version of the layout. Assuming X is given first in this code, the numbers need to be switched around (this is due to a 90° rotation in the detector chips since I made the first version of the document).

lpsinger:
> X is the larger gap size and Y the smaller one in this version of the layout. Assuming X is given first in this code, the numbers need to be switched around (this is due to a 90° rotation in the detector chips since I made the first version of the document).

Understood. Please check again.

hpearnshaw (approved):

PR #547, opened 2026-08-18
Speed up solve_tsp with some added cuts

lpsinger:
- Explicitly eliminate 1-cycles (loops).
- Explicitly eliminate 2-cycles.

These cuts were suggested by OpenAI ChatGPT 5.3 Codex via Continue. They are found to dramatically improve solution speed for UVEX sky blocks slew patterns.

### DIFF
--- a/src/m4opt/utils/optimization.py
+++ b/src/m4opt/utils/optimization.py
@@ -295,20 +295,20 @@ def solve_tsp(distances: np.ndarray, **kwargs) -> tuple[np.ndarray, float]:
     n = len(distances)
     assert n >= 2
     with Model(**kwargs) as m:
-        x = m.binary_vars((n, n))
+        # Diagonal ements of x must be zero
+        x = m.binary_vars((n, n), ub=1 - np.eye(n))
         y = m.integer_vars(n - 1, lb=1, ub=n - 1)
-        m.add_constraints_(
-            [
-                m.sum_vars_all_different([x[i, j] for i in range(n) if i != j]) == 1
-                for j in range(n)
-            ]
-        )
-        m.add_constraints_(
-            [
-                m.sum_vars_all_different([x[i, j] for j in range(n) if i != j]) == 1
-                for i in range(n)
-            ]
-        )
+
+        # Exactly one edge is taken into and out of each node.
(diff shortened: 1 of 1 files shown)


src/m4opt/utils/optimization.py
```diff
@@ -317,16 +316,7 @@ def solve_tsp(distances: np.ndarray, **kwargs) -> tuple[np.ndarray, float]:
...
-            m.sum(
-                [
-                    distances[i, j] * x[i, j]
-                    for i in range(n)
-                    for j in range(n)
-                    if i != j
-                ]
-            )
-        )
+        m.minimize(m.scal_prod_vars_all_different(x.ravel(), distances.ravel()))
```
mcoughlin, line 320: You can keep the speed-up of ignoring the diagonal with something like:

coeffs = np.where(np.eye(n, dtype=bool), 0.0, distances)
m.minimize(m.scal_prod_vars_all_different(x.ravel(), coeffs.ravel()))

src/m4opt/utils/optimization.py
```diff
@@ -295,20 +295,19 @@ def solve_tsp(distances: np.ndarray, **kwargs) -> tuple[np.ndarray, float]:
     n = len(distances)
     assert n >= 2
     with Model(**kwargs) as m:
-        x = m.binary_vars((n, n))
+        x = m.binary_vars((n, n), ub=1 - np.eye(n))
```
mcoughlin, line 299: Add a brief comment? The expression is a bit weird before thinking about it :)

mcoughlin (changes requested):
A couple of comments

src/m4opt/utils/optimization.py
lpsinger: This probably wouldn't change the coefficients because the diagonal elements of `distances` had better be zero anyway. Even if they weren't, it wouldn't seed it up because the number of terms would be the same (same number of elements of `x.ravel()`).

src/m4opt/utils/optimization.py
lpsinger: Done.

mcoughlin (approved):
Thanks @lpsinger. LGTM

PR #551, opened 2026-08-19
Raise NotImplementedError in all placeholder abstract methods

lpsinger:
In Python, a class must implement all of its bases abstract methods in order to be instantiated, but the base class abstract methods may still be called. This is a potential footgun (see #549).


mcoughlin (approved):
LGTM

PR #553, opened 2026-08-21
Fix footprint and footprint_healpix for empty compound Regions

lpsinger:
Both crashed on an empty Regions([]): footprint_inner called ArrayOfRegions() with no args (it requires at least one), and concat_healpix concatenated zero arrays. Handle the empty case explicitly in each and document it with doctest examples.


mcoughlin (approved):
LGTM

PR #554, opened 2026-08-21
Add AntiSolarSeparationConstraint

Eliza-Diggins:
Added a new constraint (``AntiSolarSeparationConstraint``) to the ``m4opt.constraints`` module which can be used to prevent instruments from targeting the anti-sun direction. 

## Justification / Discussion

As noted in [m4opt/uvex-scheduler#30](https://github.com/m4opt/uvex-scheduler/issues/30), the nominal roll field $\boldsymbol{\psi}(\hat{\bf n}, t) \in T(S^2)$ is not smooth at the sun and anti-sun positions. In the [UVEX scheduler](https://github.com/m4opt/uvex-scheduler), this can result in long slews between adjacent fields within a single observing block if that block contains either the sun or anti-sun points at its scheduled time. Currently, ``m4opt.missions._uvex`` prevents the instrument from pointing at the sun, but does not prevent it from pointing at the anti-sun point. 

This PR simply implements an ``AntiSolarSeparationConstraint`` which can be added to the UVEX mission (not done here) to prevent this pathology from arising. 

## Notes

Unlike the ``SunSeparationConstraint``, which requires a reasonably large solid angle of exclusion, the anti-sun separation constraint may be treated as effectively point-like (the field is everywhere else smooth), and so the ``min`` parameter can be made quite small.

### DIFF
--- a/src/m4opt/constraints/_body_separation.py
+++ b/src/m4opt/constraints/_body_separation.py
@@ -88,42 +88,39 @@ class SunSeparationConstraint(BodySeparationConstraint):
         super().__init__(min, "sun")
 
 
-class AntiSolarSeparationConstraint(BodySeparationConstraint):
-    def __init__(self, min: u.Quantity[u.physical.angle]):
-        """
-        Constrain the minimum separation from the anti-solar point.
+class AntiSolarSeparationConstraint(SunSeparationConstraint):
+    """
+    Constrain the minimum separation from the anti-solar point.
 
-        The anti-solar point is the point on the sky directly opposite the
-        Sun (solar elongation of 180°). The nominal spacecraft roll angle
-        (see :func:`~m4opt.dynamics.nominal_roll`) is undefined there, and
-        changes arbitrarily fast nearby, so it is generally necessary to
-        keep targets away from it by some margin.
+    The anti-solar point is the point on the sky directly opposite the
+    Sun. The nominal spacecraft roll angle
+    (see :func:`~m4opt.dynamics.nominal_roll`) is undefined there, and
+    changes arbitrarily fast nearby.
 
-        Parameters
-        ----------
-        min
-            Minimum angular separation from the anti-solar point.
+    Parameters
+    ----------
+    min
+        Minimum angular separation from the anti-solar point.
 
-        See Also
-        --------
-        m4opt.constraints.SunSeparationConstraint
-        m4opt.dynamics.nominal_roll
+    See Also
+    --------
+    m4opt.constraints.SunSeparationConstraint
+    m4opt.dynamics.nominal_roll
 
-        Examples
-        --------
+    Examples
+    --------
 
-        >>> from astropy.coordinates import EarthLocation, SkyCoord
-        >>> from astropy.time import Time
-        >>> from astropy import units as u
-        >>> from m4opt.constraints import AntiSolarSeparationConstraint
-        >>> time = Time("2017-08-17T12:41:04Z")
-        >>> target = SkyCoord.from_name("NGC 4993")
-        >>> location = EarthLocation.of_site("Las Campanas Observatory")
-        >>> constraint = AntiSolarSeparationConstraint(20 * u.deg)
-        >>> constraint(location, target, time)
-        np.True_
-        """
-        super().__init__(min, "sun")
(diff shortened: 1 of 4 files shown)


Eliza-Diggins:
@lpsinger, I haven't attached the ``AntiSolarSeparationConstraint`` to the UVEX mission, but can do pending your approval.

bwgref:
Why are you not allowing UVEX to point in the anti-Sun direction? This is a perfectly fine orientation.

lpsinger:
> Why are you not allowing UVEX to point in the anti-Sun direction? This is a perfectly fine orientation.

There are two problems with the anti-Sun direction:
- Gegenschein, enhanced zodiacal light
- Nominal roll angle changes very rapidly with sky location, leads to nearly 180° slews even for nearest-neighbor fields

lpsinger (changes requested):
This is a great idea. And implementing the constraint doesn't mean we have to use it. It just adds it to our toolbox.

We can reduce code duplication in one of two ways:

- Define a `get_body` method on `BodySeparationConstraint`, and override it to invert the Cartesian coordinates, or
- Define a `_separation` method in `BodySeparationConstraint` that return `get_body(...).separation(...)`. Then you can override that method in your class to `return 180 * u.deg - super()._separation(...)`.

bwgref:
"Constraint" sounds like "We can't do X".

If there are performance metrics (like enhanced Zodi or long slews), that's fine. But I'd argue against labeling it as a constraint just in case someone downstream reads it as "Oh, there's something about the UVEX spacecraft that *can't* do this, ergo there's big hole in the field of regard anti-Sun". Which is wrong.

lpsinger:
> "Constraint" sounds like "We can't do X".

The `m4opt.constraints` module contains a large collection of constraints that can be composed together. This PR does not represent a decision or even a proposal to impose this constraint in any operational context. However it adds it to our potential toolbox.

All of these classes have a name that ends in Constraint. I don't propose to change that pattern.

src/m4opt/constraints/_body_separation.py
```diff
@@ -85,3 +86,47 @@ def __init__(self, min: u.Quantity[u.physical.angle]):
...
         super().__init__(min, "sun")
+
+
+class AntiSolarSeparationConstraint(BodySeparationConstraint):
+    def __init__(self, min: u.Quantity[u.physical.angle]):
+        """
+        Constrain the minimum separation from the anti-solar point.
+
+        The anti-solar point is the point on the sky directly opposite the
+        Sun (solar elongation of 180°). The nominal spacecraft roll angle
```
lpsinger:
```suggestion
        Sun. The nominal spacecraft roll angle
```

The solar elongation is not enough to define the anti-solar point.

src/m4opt/constraints/_body_separation.py
```diff
@@ -85,3 +86,47 @@ def __init__(self, min: u.Quantity[u.physical.angle]):
...
+        (see :func:`~m4opt.dynamics.nominal_roll`) is undefined there, and
+        changes arbitrarily fast nearby, so it is generally necessary to
+        keep targets away from it by some margin.
```
lpsinger:
```suggestion
        changes arbitrarily fast nearby.
```

src/m4opt/constraints/_body_separation.py
```diff
@@ -85,3 +86,47 @@ def __init__(self, min: u.Quantity[u.physical.angle]):
         np.True_
         """
         super().__init__(min, "sun")
+
+
+class AntiSolarSeparationConstraint(BodySeparationConstraint):
```
lpsinger:
```suggestion
class AntiSolarSeparationConstraint(SunSeparationConstraint):
```

Also move the docstring from the `__init__` method to the class itself (see #557). Then you can delete the `__init__` method entirely.

lpsinger (changes requested):

lpsinger (approved):
:clap: very nice work! This is a clever addition.

lpsinger:
When I merged #557 it introduced a merge conflict. Please fix.

Eliza-Diggins:
> When I merged #557 it introduced a merge conflict. Please fix.

Resolved!

lpsinger (approved):

PR #556, opened 2026-08-22
Add the `m4opt.models` subpackage: spectral modeling for synthetic photometry

Eliza-Diggins:
## Summary

This PR adds `m4opt.models`, the infrastructure for constructing source models used by `m4opt.synphot`'s exposure-time/SNR calculations and for end-to-end simulations of instrument observations. It provides a small set of composable abstractions for spectral shapes, bolometric light curves, and full spectral energy distributions, together with a curated library of concrete transient models.

Every model is **probabilistic**: each physical parameter carries a prior distribution, so a model can be evaluated at fixed, literal parameter values or sampled to produce a whole population of simulated sources — e.g. to test how sensitive a survey strategy is to the diversity of a transient population, rather than to one fiducial light curve.

See `docs/guide/models.rst` for the full guide; the highlights are below.

## Core abstractions (`m4opt.models.core`)

A spectral model is a rule for the spectral luminosity $L_\nu(\nu, t | \boldsymbol{\Theta})$ as a function of observer-frame frequency, time since a reference epoch, and a set of named physical parameters. Models come in three flavors depending on what they vary with:

- **`Spectrum`** — a spectral shape that depends on frequency only (e.g. a blackbody).
- **`Lightcurve`** — a bolometric light curve that depends on time only.
- **`SpectralModel`** — a full SED that depends on both frequency and time; most users only ever work with concrete `SpectralModel` subclasses, with `Spectrum`/`Lightcurve` as the building blocks a `SpectralModel` is assembled from.

Each named parameter is backed by a `Parameter`: a `Prior` to sample from, a characteristic physical `scale`, and an optional `transform` (e.g. `"log10"`) that lets the prior be defined on a more convenient variable than the physical value itself. Parameters can be overridden at construction time either by replacing the prior entirely or by pinning ("fixing") a parameter to one specific value.

Every quantity a `SpectralModel` can compute is available through a consistent family of classmethods, distinguished by suffix (`_log_cgs`, `_log`, `_cgs`, or unit-aware with no suffix), split into rest-frame/undiluted quantities (`eval`, `eval_bolometric`, `eval_spectrum`) and observed/diluted quantities that account for redshift and distance (`flux`, `mag`, and their band-integrated `*_band` counterparts).

## Model library

- **`m4opt.models.spectra`** — basic spectral shapes: `BlackbodySpectrum`, `PowerLawSpectrum`, `BrokenPowerLawSpectrum`.
- **`m4opt.models.lightcurves`** — generic light curve shapes: `FREDLightcurve` (fast-rise, exponential-decay), `BazinLightcurve`, `VillarLightcurve`, and others.
- **`m4opt.models.supernovae`** — composite supernova models: `VillarCoolingBlackbodySED`.
- **`m4opt.models.tdes`** — composite tidal disruption event models: `VanVelzenTDESED`.

Every class is importable directly from `m4opt.models`, and each model's docstring includes a parameter table listing every parameter's symbol, description, and default (physically-motivated) prior.

## Simulating observations

The intended end-to-end workflow is entirely NumPy-broadcasting based, so it vectorizes for free across whole populations of sources rather than looping one light curve at a time:

1. sample or fix a model's parameters per simulated source
2. pick observation times, pointings, source redshifts, and distances.
3. Generate the **vectorized** `SourceSpectrum`` object. 
4. feed that to `m4opt.synphot.Detector`

`docs/guide/models.rst` walks through a fully worked example: simulating a UVEX survey of several core-collapse supernovae, including field-of-regard constraints, and plotting the resulting sampled light curves against the underlying continuous model.

## Extensibility

The module is designed to be easily extended: `docs/guide/models.rst` (Building Custom Models / Developer Notes) documents the contracts for subclassing `Prior`, `Parameter`, and the `Spectrum`/`Lightcurve`/`SpectralModel` base classes, including the composite-model pattern (`ComposedSpectralModel`) used by the built-in supernova/TDE models to combine a `Lightcurve` and a `Spectrum`.

## Testing

- A full contract-based test suite (`m4opt/models/tests/`) exercises every concrete `Spectrum`/`Lightcurve`/SED subclass against a shared set of correctness properties (broadcasting, unit consistency, normalization, sampling), plus dedicated synthetic-photometry round-trip tests against `m4opt.synphot`.
- All docstring and `docs/` examples are collected and run as doctests.
- `mypy`, `pre-commit` (ruff/codespell), and a full `sphinx-build -W` docs build are all clean.

## Known limitations

- **Multi-axis batching through `as_source_spectrum` requires manually-reserved axes.** Batching a single source across multiple times already worked without any special care. But batching *both* events and time together (the pattern a real scheduling run uses) requires the caller to manually reserve a distinct trailing axis for each batch dimension on `t` and on every batched parameter, so they don't collide with the wavelength axis `synphot` samples internally: `as_source_spectrum` does not insert these axes for you. Get the axis order wrong and it broadcasts silently, not with an error. See `test_synthetic_photometry_batches_events_and_times` for the working pattern.

  @lpsinger, I think the ergonomics of this can be improved, but would require me touching ``m4opt.synphot``, which may require some care to ensure backward compatibility.

- **Dust extinction is not compatible with a batched `SourceSpectrum`.** `m4opt.synphot`'s countrate interpolation shortcut for dust extinction assumes the source spectrum carries no batch axis of its own, and breaks when a per-event batched SED is used. Tracked as an `xfail` (`test_synthetic_photometry_with_dust_extinction`): needs a fix in `_math.py` to interpolate along a trailing `Ebv` axis instead of assuming a scalar result. Until then, applying dust extinction to a batch of sources requires iterating one at a time.

  @lpsinger, Not sure how we want to handle this one. Could modify the ``Ebv`` system to handle the trailing axes properly I think...

## Test plan

- [x] `pytest --doctest-plus --pyargs m4opt docs` passes (645 passed, 7 skipped — CPLEX Community Edition size limits, unrelated — 1 xfailed)
- [x] `mypy` clean on `m4opt.models`
- [x] `pre-commit run --all-files` clean
- [x] `sphinx-build -W -b html docs docs/_build/html` clean (models-related content)


lpsinger (changes requested):
This is a gigantic PR. It adds:

1. A completely new parametric spectral model abstraction that duplicates much of the prior art of [astropy.modeling](https://docs.astropy.org/en/stable/modeling/index.html) or [synphot](https://synphot.readthedocs.io/en/latest/)
2. A Monte Carlo sampling framework that is reminiscent of, but completely independent from, [astropy.uncertainty](https://docs.astropy.org/en/stable/uncertainty/index.html) or [scipy.stats](https://docs.scipy.org/doc/scipy/reference/stats.html)
3. Low-value wrappers and abstractions of already easy to use things like [astropy.cosmology](https://docs.astropy.org/en/stable/cosmology/index.html)
4. The actual models themselves.

I don't want to reinvent wheels 1-3. Integrating spectral modeling with automated Monte Carlo sampling could be handy, but there is surely a way to do this that enhances the entire existing astropy.modeling ecosystem (of which synphot is a part). But let's decouple that from the transient light curve / SED models.

Remember that it will ultimately be my task to integrate the new models with the accelerated synthetic photometry in `m4opt.synphot`. That would be an almost insurmountable task with all of this other new abstraction is in the mix.

In this PR, would you simply provide item 4 as a [collection of synphot-style Astropy model subclasses](https://synphot.readthedocs.io/en/latest/#id5) with [Astropy model parameters](https://docs.astropy.org/en/latest/modeling/parameters.html)?

By the bye, I think that `m4opt.synphot.source` would be a better place for models of source SEDs.

Eliza-Diggins:
> This is a gigantic PR. It adds:
> 
> 1. A completely new parametric spectral model abstraction that duplicates much of the prior art of [astropy.modeling](https://docs.astropy.org/en/stable/modeling/index.html) or [synphot](https://synphot.readthedocs.io/en/latest/)
> 2. A Monte Carlo sampling framework that is reminiscent of, but completely independent from, [astropy.uncertainty](https://docs.astropy.org/en/stable/uncertainty/index.html) or [scipy.stats](https://docs.scipy.org/doc/scipy/reference/stats.html)
> 3. Low-value wrappers and abstractions of already easy to use things like [astropy.cosmology](https://docs.astropy.org/en/stable/cosmology/index.html)
> 4. The actual models themselves.
> 
> I don't want to reinvent wheels 1-3. Integrating spectral modeling with automated Monte Carlo sampling could be handy, but there is surely a way to do this that enhances the entire existing astropy.modeling ecosystem (of which synphot is a part). But let's decouple that from the transient light curve / SED models.

Hey @lpsinger, I totally agree. I didn't originally write this code with ``m4opt`` in mind, it was only meant to be a quick and dirty infrastructure to allow me to perform the transient simulations we want. What you see here is my best attempt at porting that infrastructure without starting from scratch. I do think that an ``astropy.model`` based approach is cleaner in many respects (this is probably what I would have done had I been building from scratch for ``m4opt``). 

There would, nonetheless, be some pinch points:

- Even with some of the boilerplate replaced with ``astropy.model.Model``, there is a fundamental issue which is that ``synphot`` doesn't appear to be well designed for vector-valued parameters / batching over many realizations of the same / similar spectra. My workaround (to get operability with your ``observing`` context) was to define parameter-free models so that we didn't trip any errors on the ``synphot`` end. Perhaps you know a better approach? I was not able to make that functional.
- The dust extinction infrastructure is not designed to work with batched evaluation of models. This requires modifications to ``_math.py``. 

Either way, I'm happy to build that infrastructure, but I'm not sure if it should be our first priority (UVEX wise. ``m4opt`` wise I'm totally with you)? This code is functional and reasonably fast at scale. I worry that the development time required to build everything around a new underlying paradigm will place me behind the 8-ball in terms of actually delivering on the science.

Eliza-Diggins:
@lpsinger, here's the problematic code from ``synphot`` [took me a while to track down]:

``` python
# TODO: Update model logic when astropy.modeling supports Quantity.
class BaseSpectrum:
    """Base class to handle spectrum or bandpass.

    .. note::

        Until `astropy.modeling` can handle units, all parameters
        are converted to pre-defined internal units.

    Parameters
    ----------
    modelclass : cls
        Model class from `astropy.modeling`.

    clean_meta : bool
        Scrub "expr" and "header" entries from input metadata before merging.
        Set this to `True` when those entries no longer make sense in ``self``.
        This is automatically set to `True` regardless for spectrum arithmetic.

    kwargs : dict
        Model parameters accepted by ``modelclass``. Each parameter can
        be either a Quantity or number. If the latter, assume pre-defined
        internal units.

    Attributes
    ----------
    meta : dict
        Metadata associated with the spectrum or bandpass model \
        (warnings, legacy SYNPHOT expression, FITS header, etc).

    Raises
    ------
    synphot.exceptions.SynphotError
        Invalid model.

    """
    ...

    def __init__(self, modelclass, clean_meta=False, **kwargs):

        # Does not handle multiple model sets for now; too complicated.
        n_models = kwargs.pop('n_models', 1)
        if n_models != 1:
            raise exceptions.SynphotError('Model can only have n_models=1')
```

So if we use an ``astropy.modeling.Model`` subclass to represent the spectrum, it **cannot** use ``Parameter`` declarations because it will break compatibility with ``synphot``. 

A few things that we can do:

1. **Remove Priors**: You're right that these could be removed. In practice ``astropy.uncertainty`` provides these for free and we can just rely on that contract.
2. **Remove Cosmology Protections**: The cosmology wrappers were written for convenience when doing things like deducing $z$ from $D_L$ or $D_A$, but we can remove them and just require that users either explicitly provide $D_L$ and $z$ separately or that they provide a $z$ and a cosmology object.

I don't think we can remove the dynamical generation of the ``SourceSpectrum`` so easily. If we subclass ``astropy.modeling.Model`` we still have to come up with a work-around for the 0-D parameter requirement from ``synphot`` which we currently handle by baking the parameters into the class declaration. 

A lot of the code in the PR is for computing fluxes / luminosities / magnitudes, which can all be delegated to ``SourceSpectrum``. This is a viable approach with the downside that we have to invoke a set of ``**parameters`` before being able to compute spectra / lightcurves instead of having a generic ``eval(nu, t, **parameters)`` available.

lpsinger:
> * Even with some of the boilerplate replaced with `astropy.model.Model`, there is a fundamental issue which is that `synphot` doesn't appear to be well designed for vector-valued parameters / batching over many realizations of the same / similar spectra. My workaround (to get operability with your `observing` context) was to define parameter-free models so that we didn't trip any errors on the `synphot` end. Perhaps you know a better approach? I was not able to make that functional.
> * The dust extinction infrastructure is not designed to work with batched evaluation of models. This requires modifications to `_math.py`.

Like I said when we met on Zoom I am working on generalizing M4OPT's accelerated synthetic photometry code to vectorize over arbitrary parameters. So the most useful thing would be to provide models that have ordinary parameters.

Please, if you are using AI to generate PR comments, don't just copy and paste responses. Take some time to make them your own and make them reasonably terse, because a human is reading and reviewing them. This project ascribes to [Astropy's AI policy](https://github.com/astropy/astropy-project/blob/main/policies/ai-policy.md), namely:

> Authentic engagement. The pull request process is collaborative and iterative. Contributors are expected to actively engage with reviewer feedback themselves. Copying and pasting replies to / from a generative AI tool does not count as engaging with the reviewer; contributors are expected to understand the changes they are proposing, not act as message courier for a third party.

Eliza-Diggins:
> > * Even with some of the boilerplate replaced with `astropy.model.Model`, there is a fundamental issue which is that `synphot` doesn't appear to be well designed for vector-valued parameters / batching over many realizations of the same / similar spectra. My workaround (to get operability with your `observing` context) was to define parameter-free models so that we didn't trip any errors on the `synphot` end. Perhaps you know a better approach? I was not able to make that functional.
> > * The dust extinction infrastructure is not designed to work with batched evaluation of models. This requires modifications to `_math.py`.
> 
> Like I said when we met on Zoom I am working on generalizing M4OPT's accelerated synthetic photometry code to vectorize over arbitrary parameters. So the most useful thing would be to provide models that have ordinary parameters.

Okay, I can do that. To ensure that we're on the same page, is your vision that we have some version of the following:

```python
from astropy.modeling import Model

class SpectralModel(Model):

    # Some parameter declarations plugging into astropy's uncertainty framework.

    # Then some utility methods for evaluating the model
```

How do you want that to interface with ``synphot`` (the actual library, not ``m4opt.synphot``)? Do you want me to find a way to convert to ``synphot.SourceSpectrum`` or do you want to take over at the ``astropy.modeling.Model`` level?

What is the minimal contract you want in order to be able to wrap things into your synthetic photometry system?

> Please, if you are using AI to generate PR comments, don't just copy and paste responses. Take some time to make them your own and make them reasonably terse, because a human is reading and reviewing them. This project ascribes to [Astropy's AI policy](https://github.com/astropy/astropy-project/blob/main/policies/ai-policy.md), namely:
> 
> > Authentic engagement. The pull request process is collaborative and iterative. Contributors are expected to actively engage with reviewer feedback themselves. Copying and pasting replies to / from a generative AI tool does not count as engaging with the reviewer; contributors are expected to understand the changes they are proposing, not act as message courier for a third party.

To be *very clear*, none of the comments on this thread have been AI generated. Perhaps I just tend to over explain? I let Claude generate the PR doc because (as you noted) it was a large piece of code and I wanted to be comprehensive in my description without wasting a lot of time. But I don't understand why you think these PR comments are AI generated?

Eliza-Diggins:
Closing in favor of #558, which I think is more in line with what we want.

lpsinger:
> To be _very clear_, none of the comments on this thread have been AI generated. Perhaps I just tend to over explain? I let Claude generate the PR doc because (as you noted) it was a large piece of code and I wanted to be comprehensive in my description without wasting a lot of time. But I don't understand why you think these PR comments are AI generated?

I'm sorry about that. I made an unwarranted assumption because of how quick and detailed your responses were. Keep that up!

PR #557, opened 2026-08-23
Move docstrings from __init__ method to class

lpsinger:
All other classes are like that. I don't know why I put these docstrings on the __init__ methods.


mcoughlin (approved):
LGTM

PR #558, opened 2026-08-24
Adding Source Models to synphot

Eliza-Diggins:
## Summary

A modified approach to #556 which builds on ``astropy.modeling.Model`` instead of an out-of-the-box modeling structure. This PR introduces the following:

- ``m4opt.synphot.model._prior``: Utilizes ``scipy.stats`` distributions to provide unit-bearing prior distributions. *Note*: ``astropy.uncertainty`` cannot provide this infrastructure out of the box because distributions cannot be instantiated as abstract objects.
- ``m4opt.synphot.model._parameter``: Overwrites the ``astropy.modeling.Parameter`` class with a ``PriorParameter`` class which is a thin wrapper around the base class with simple infrastructure for assigning a prior.
- ``m4opt.synphot.model._base``: The core of the new module. Provides a mixin class for ``astropy.modeling.Model`` (``_ModelSamplingMixin``) which allows astropy models to sample from their parameter priors. Descended from these are abstract ``SEDModel``, ``Lightcurve`` and ``Spectrum`` models. A ``ComposedSEDModel`` is also included.

In addition, ``_spectra.py``, ``_lightcurves.py``, ``_tdes.py`` and ``_supernovae.py`` implement concrete examples of these models.

## To-Do / Discussion

- [ ] At present, ``SEDModel`` only provides its ``evaluate`` method as a proxy for $$L_\nu(\nu,t)$$. How this interfaces with the rest of ``m4opt.synphot`` (or for that matter ``synphot`` itself) remains to be seen. Batched instantiation of ``SEDModel`` is not compatible with ``synphot.SourceSpectrum`` by construction.
- [ ] No testing present yet.
- [ ] No documentation present yet.


Eliza-Diggins:
@lpsinger , go ahead and have a look at this. It's not comprehensive yet (no testing, no docs, probably some more infrastructure to build), but I want to get your go ahead on the basic structure before proceeding.

lpsinger (changes requested):
- I'd prefer the module name `m4opt.synphot.source`, to mirror the existing `m4opt.synphot.background` module. There are _many_ types of models in this project; not all of the models are synthetic photometry models. The `m4opt.synphot` module is for synthetic photometry models.
- Rather than model an achromatic model as a class whose `evaluate` method multiplies the time dependence and the SED dependence, you should implement it as a product of two model objects. The time dependence is a frequency-independent factor that my integration machinery will easily be able to vectorize over.
- Subclassing Parameter to add a sampling distribution is an interesting idea, but it is not clear if all of these parameters do have one canonical distribution. Will it ever be desirable to be able to modify the distributions of these parameters? Are these distributions simply the ones that you have chosen for your simulation? If so, then it would make more sense to be able to configure the distributions separately.

Eliza-Diggins:
Hey @lpsinger:

> I'd prefer the module name m4opt.synphot.source, to mirror the existing m4opt.synphot.background module. There are many types of models in this project; not all of the models are synthetic photometry models. The m4opt.synphot module is for synthetic photometry models.

That makes sense. I'll refactor to ``m4opt.synphot.source``.

> Rather than model an achromatic model as a class whose evaluate method multiplies the time dependence and the SED dependence, you should implement it as a product of two model objects. The time dependence is a frequency-independent factor that my integration machinery will easily be able to vectorize over.

I think this should be easy enough! I'll have to look a little at how ``astropy.modeling.Model`` constrains the behavior of the ``__mul__`` dunder, but this seems doable... I guess the other question is whether ``__mul__`` can be a classmethod or not. But if not I can probably define a metaclass to work around it. If not, we just force users to instantiate (which they probably do anyways) and just overwrite ``__mul__`` (or maybe it works out of the box? I'll have to check).

> Subclassing Parameter to add a sampling distribution is an interesting idea, but it is not clear if all of these parameters do have one canonical distribution. Will it ever be desirable to be able to modify the distributions of these parameters? Are these distributions simply the ones that you have chosen for your simulation? If so, then it would make more sense to be able to configure the distributions separately.

Ah, this touches on an element of ``astropy.modeling.Model`` that I did not know about until I wrote this. When you instantiate a model class without explicit parameter values

```python
model = ModelClass()
```

the ``model.__dict__`` gets handed deep copies of the ``Parameter`` objects attached to the ``ModelClass`` which are mutable. So the design pattern I've gone with here is something like this:

```python
from m4opt.synphot.source import APhotometryModel, APriorDistribution

# The photometry model's evaluate method is available as a class method if I just want L_nu:
lnu = APhotometryModel.evaluate(1 * u.Hz, 1*u.s, *my_parameters)

# To **sample** from the model we have to actually instantiate, which hands us a copy of all the parameters that we 
# can mutate.
model = APhotometryModel()

# I don't love the default prior, so will want to change it...
my_prior = APriorDistribution(mean=..., std=..., unit=...)
model.my_parameter.prior = my_prior

# Now we can sample either directly from the prior or for all of the parameters using the ``model.sample`` method.

parameter_sample = model.sample(n=5000)
```

It's not the most intuitive contract for the user to have to instantiate to sample but not have to instantiate for access to ``evaluate``; however, I think the typical user just instantiates every class they interact with, which will work just fine. We could also make ``evaluate`` a standard method (not ``classmethod``), but that may be a limitation in the future...

lpsinger:
> I think this should be easy enough! I'll have to look a little at how `astropy.modeling.Model` constrains the behavior of the `__mul__` dunder, but this seems doable... I guess the other question is whether `__mul__` can be a classmethod or not. But if not I can probably define a metaclass to work around it. If not, we just force users to instantiate (which they probably do anyways) and just overwrite `__mul__` (or maybe it works out of the box? I'll have to check).

Astropy models already support multiplication.

Eliza-Diggins:
> > I think this should be easy enough! I'll have to look a little at how astropy.modeling.Model constrains the behavior of the __mul__ dunder, but this seems doable... I guess the other question is whether __mul__ can be a classmethod or not. But if not I can probably define a metaclass to work around it. If not, we just force users to instantiate (which they probably do anyways) and just overwrite __mul__ (or maybe it works out of the box? I'll have to check).

> Astropy models already support multiplication.

So I've done some experimentation and it seems like this will actually be quite complex to make work. The following won't raise an error

```python
from m4opt.synphot.source import BlackbodySpectrum, GREDLightcurve
bb_spec = BlackbodySpectrum()
lc = GREDLightcurve()
joint_model = bb_spec * lc
```

However, 

```python
joint_model.inputs
>>> ('freq',)
joint_model.n_inputs
>>>  1
```

Under the hood, the infrastructure behind ``CompoundModel`` assumes (with no safety check, which surprises me... maybe a PR to astropy) that the two models have the same inputs, which is not the case here. 

We're fortunate that ``joint_model.temperature_0`` is still a ``PriorParameter`` after composition of this sort, but the naming convention will be difficult for downstream users to understand and, more importantly, ``CompoundModel`` doesn't have a ``sample`` method, so after composition you lose the ability to MC sample from the model. 

I think one could attempt to circumvent this, but it would require some very fragile modifications to parts of astropy that I don't think we want to mess with. In most use cases, I think we probably don't lose that much by just biting the bullet and saying that we cannot decompose in the operation tree and we just treat everything as $L_\nu(t,\nu)$ regardless...

A broader question I hadn't given much thought to is exactly how this class structure functions through ``__mul__``, ``__div__``, etc. We don't want something like ``new_model = 5*old_model`` to suddenly change its underlying type and lose sampling...

lpsinger:
You could declare an extrinsic scale factor class (https://github.com/m4opt/m4opt/blob/main/src/m4opt/synphot/_extrinsic.py) that takes as its argument any arbitrary Astropy model that is a function of epoch.

lpsinger:
As for the sampling capability, I notice that [astropy.modeling.Parameter](https://docs.astropy.org/en/stable/api/astropy.modeling.Parameter.html) has both a `prior` and a `posterior` attribute, although they don't seem to be used for much. It may be worth asking on the Astropy Slack what these are for and how they are used.

Eliza-Diggins:
> You could declare an extrinsic scale factor class (https://github.com/m4opt/m4opt/blob/main/src/m4opt/synphot/_extrinsic.py) that takes as its argument any arbitrary Astropy model that is a function of epoch.

That could work! I suppose the pattern would need to be defining an ``ExtrinsicLightcurveModel(ABC, ExtrinsicScaleFactor, _ModelSamplingMixin)``. I think that this will solve the composition issue; however, it brings up a few other considerations:

1. If we adopt this approach, then I can do the following:

```python
from m4opt.synphot.source import MySpectrum, MyLightcurve

model = MySpectrum() * MyLightcurve()
```

however, this ``model`` object is now ``CompoundModel`` and doesn't have the Monte Carlo sampling. Thus,

```
model.sample()
```

would generate an error. I wonder if this would produce a weird dichotomy in behavior between the true SEDs and the composed models since one would want (intuitively) to have the same capabilities on both classes.

2. A smaller consideration is that ``ExtrinsicScaleFactor`` has access to the observing ``state`` and therefore to the observation time, but current models are defined without a defined ``t_0``. This is easily worked around: we would need to make the instantiation contract ``MyModel(t_0: Time)`` and then the ``ExtrinsicScaleFactor.value`` attribute would pass ``t = t_obs - t_0`` into the model's ``evaluate`` method. But this does then mean that our notion of a "spectral model" is intrinsically linked to a start time, which may be unintuitive...


> As for the sampling capability, I notice that [astropy.modeling.Parameter](https://docs.astropy.org/en/stable/api/astropy.modeling.Parameter.html) has both a `prior` and a `posterior` attribute, although they don't seem to be used for much. It may be worth asking on the Astropy Slack what these are for and how they are used.

As far as I can tell they're completely unused at present. I'm not in the astropy slack, but if you want to check in and ask, I'd be interested to know if there's a reason for the class to have unused attributes.

lpsinger:
> I'm not in the astropy slack, but if you want to check in and ask, I'd be interested to know if there's a reason for the class to have unused attributes.

You can join. It's open to all.

PR #563, opened 2026-08-31
Remove unnecessary MANIFEST.in

lpsinger:
This is archaic and unnecessary because modern setuptools puts everything that is under version control into the sdist.


mcoughlin (approved):
LGTM

PR #565, opened 2026-08-31
ULTRASAT's tabulated throughput curve

mcoughlin:
This PR uses ULTRASAT's tabulated throughput curve.


lpsinger:
How heavy of a dependency would sncosmo itself be? That way we wouldn’t have to keep this file up to date.

mcoughlin:
@lpsinger The issue is that the sncosmo fetch is 4.4 MB (ULTRASAT_TR.dat, with all 25 radial columns) versus our 56 KB vendored on-axis table. My experience with sncosmo in CI is very bad generally (I maintain a "skyportal-data") repo for this reason...

lpsinger (changes requested):
Is sncosmo the primary source, or do they fetch and vendor it from somewhere else?

Would you add a unit test that includes your code to generate the downsampled file and check that it matches exactly? There may even be a pytest plugin for comparing an output with a saved specimen, that with appropriate flags to pytest would regenerate it.

mcoughlin:
@lpsinger  I got them from Yossi a long time ago and put them here: https://github.com/sncosmo/sncosmo.github.io/tree/master/data/bandpasses/ultrasat

I am adding the comparison to your weekly CI.

src/m4opt/missions/_ultrasat/tests/test_throughput.py
```diff
@@ -0,0 +1,77 @@
...
+    split = np.searchsorted(wavelength, FINE_SAMPLING_BELOW)
+    keep = np.unique(
+        np.concatenate(
+            [
+                np.arange(0, split),
+                np.arange(split, len(wavelength), COARSE_SAMPLING),
+                [len(wavelength) - 1],
+            ]
+        )
+    )
```
lpsinger: As constructed, these indices are already unique.

```suggestion
    keep = np.concatenate(
            [
                np.arange(0, split),
                np.arange(split, len(wavelength) - 1, COARSE_SAMPLING),
                [len(wavelength) - 1],
            ]
        )
```

src/m4opt/missions/_ultrasat/tests/test_throughput.py
```diff
@@ -0,0 +1,77 @@
...
+
+@pytest.mark.remote_data
+def test_throughput_matches_upstream():
+    """The vendored table is the on-axis column of the upstream ULTRASAT data.
+
+    Regenerate it by writing the result of :func:`downsample` over
+    ``data/throughput.ecsv`` in ``ascii.ecsv`` format.
+    """
+    wavelength = (
+        np.loadtxt(download_file(UPSTREAM + "Wavelength.dat", cache=True)) * u.angstrom
```
lpsinger:
```suggestion
        np.loadtxt(download_file(f"{UPSTREAM}Wavelength.dat", cache=True)) * u.angstrom
```

src/m4opt/missions/_ultrasat/tests/test_throughput.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+    # One column per radial distance from the optical axis; the first is on axis.
+    transmission = np.loadtxt(
+        download_file(UPSTREAM + "ULTRASAT_TR.dat", cache=True), delimiter=","
```
lpsinger:
```suggestion
        download_file(f"{UPSTREAM}ULTRASAT_TR.dat", cache=True), delimiter=","
```

lpsinger (changes requested):

lpsinger (approved):

.github/workflows/ci_cron_weekly.yml
```diff
@@ -48,4 +55,4 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install tox
     - name: Test with tox
-      run: tox -e ${{ matrix.toxenv }}
+      run: tox -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
```
lpsinger: Rather than inject the argument this way, please add this factor to the `envlist` section in `tox.ini`:

```ini
[tox]
envlist =
    py312-test{,-alldeps,-devdeps,-oldestdeps}{,-remotedata}{,-cov}
    build_docs
    linkcheck
    codestyle
```

lpsinger (changes requested):

.github/workflows/ci_cron_weekly.yml
mcoughlin: @lpsinger fixed.

.github/workflows/ci_cron_weekly.yml
```diff
@@ -31,8 +31,7 @@ jobs:
           # depend on data hosted elsewhere, which may change or go away.
           - name: Python 3.12 with remote data
             python: '3.12'
-            toxenv: py312-test
-            toxposargs: --remote-data
+            toxenv: py312-test-remotedata
```
lpsinger: Instead of adding another test, would you please instead add them to the -alldeps-cov job?

lpsinger (changes requested):

.github/workflows/ci_cron_weekly.yml
mcoughlin: @lpsinger done. this does take it out of weekly.

lpsinger (approved):

lpsinger (changes requested):
The tests are failing.

mcoughlin:
@lpsinger sorry fixed.

lpsinger (approved):

PR #570, opened 2026-09-05
Interpolate dust extinction in log space

mcoughlin:
This PR interpolates dust extinction in log space (I was seeing differences from direct integration at the 2.3×10⁻² level; this brings it to 1.4×10⁻¹⁰).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,16 @@
 Changes
 *******
 
+2.14.0 (unreleased)
+===================
+
+- Fix the interpolation of dust extinction. The count rate was sampled on a
+  grid spaced linearly out to the most reddened pixel of the whole dust map,
+  E(B-V) of 165, so that its first step landed at 0.32 and 82% of the sky fell
+  inside it. The grid is now spaced in asinh and the logarithm of the count
+  rate is interpolated, which extinction makes nearly straight, reducing the
+  error from a few percent to about one part in a billion.
+
 2.13.0 (2026-09-08)
 ===================
 
--- a/src/m4opt/synphot/_math.py
+++ b/src/m4opt/synphot/_math.py
@@ -41,19 +41,19 @@ def countrate(
     <Quantity 1.12409479e+09 PHOTLAM>
     >>> with observing(EarthLocation.of_site('Palomar'), SkyCoord(*np.meshgrid(np.linspace(0, 360, 100), np.linspace(-90, 90, 200)), unit=u.deg), Time('2024-01-01')):
     ...     countrate(spectrum, band)
-    <Quantity [[1.15320102e+14, 1.15320102e+14, 1.15320102e+14, ...,
-                1.15320102e+14, 1.15320102e+14, 1.15320102e+14],
-               [1.25062090e+14, 1.26075744e+14, 1.27402724e+14, ...,
-                1.24497734e+14, 1.24740423e+14, 1.25062090e+14],
-               [1.32286015e+14, 1.32147384e+14, 1.31646656e+14, ...,
-                1.27972484e+14, 1.30482658e+14, 1.32286015e+14],
+    <Quantity [[1.15140438e+14, 1.15140438e+14, 1.15140438e+14, ...,
+                1.15140438e+14, 1.15140438e+14, 1.15140438e+14],
+               [1.24826976e+14, 1.25836843e+14, 1.27159526e+14, ...,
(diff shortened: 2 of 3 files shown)


lpsinger (changes requested):
This duplicates the interpolation that is already done by the accelerated synphot machinery: https://github.com/m4opt/m4opt/blob/main/src/m4opt/synphot/_math.py#L101-L113

mcoughlin:
@lpsinger My goal was to fix it, as it (i) now covers the reddenings the targets actually have, and (ii) interpolates the log of the count rate.

src/m4opt/synphot/_math.py
```diff
@@ -98,22 +98,35 @@ def base_countrate_extinction_for_Ebv(Ebv):
...
+        n_samples = 512
+        low, high = np.min(xp), np.max(xp)
+        if np.size(xp) < n_samples or not high > low:
+            return base_countrate_extinction_for_Ebv(xp) * count_rate_unit
+
+        x = np.linspace(low, high, n_samples)
+        y = base_countrate_extinction_for_Ebv(x)
+        if not np.all(y > 0):
+            # Reddening severe enough to underflow the count rate has no
+            # logarithm to interpolate.
```
lpsinger: This comment is not grammatical.

src/m4opt/synphot/_math.py
```diff
@@ -98,22 +98,35 @@ def base_countrate_extinction_for_Ebv(Ebv):
...
+            return base_countrate_extinction_for_Ebv(xp) * count_rate_unit
+
+        return (
+            np.exp(
+                interp1d(x, np.log(y), kind="cubic", copy=False, assume_sorted=True)(xp)
```
lpsinger: OK, I like the removal of the double negative in the `if:` statement, but not in this PR as it obscures what actually changed in the diff, which is only changing `interp1d(x, y, ...)` to `interp1d(x, np.log(y), ...)`.

What problem does this solve? It does introduce the underflow path which will be slow.

lpsinger (changes requested):

src/m4opt/synphot/_math.py
mcoughlin: The issue was the combination of grid range (it was np.linspace(0, Ebv_max, 512) and is now np.linspace(xp.min(), xp.max(), 512)) and the log, as while Ebv_max is 164.9, real sightlines only span about 0.0075 to 12.25.

src/m4opt/synphot/_math.py
mcoughlin: Fixed

src/m4opt/synphot/_math.py
lpsinger: By "real" sightlines, do you mean sightlines that are well away from the Galactic plane?

I think I'd rather still use a fixed interpolation grid, independent of the min and max of the passed values. What about an asinh scale?

Also please reverse the `if` statements to minimize the diff.

lpsinger (changes requested):

src/m4opt/synphot/_math.py
mcoughlin: Yeah just a couple pixels in the plane that have that crazy high Ebv_max.

lpsinger (commented):
Matplotlib's [AsinhTransform](https://github.com/matplotlib/matplotlib/blob/v3.11.1/lib/matplotlib/scale.py#L656) scales the input to make it nearly linear over a specific region. Would setting some such value help here?

mcoughlin:
@lpsinger see what you think about this version.

lpsinger (approved):

CHANGES.rst
```diff
@@ -42,6 +42,13 @@ Changes
 
 - Add ``intersect1d``, an accelerated version of ``numpy.intersect1d``.
 
+- Fix the interpolation of dust extinction. The count rate was sampled on a
+  grid spaced linearly out to the most reddened pixel of the whole dust map,
+  E(B-V) of 165, so that its first step landed at 0.32 and 82% of the sky fell
+  inside it. The grid is now spaced in asinh and the logarithm of the count
+  rate is interpolated, which extinction makes nearly straight, reducing the
+  error from a few percent to about one part in a billion.
```
lpsinger: Move to the 2.14.0 section.

lpsinger (changes requested):

CHANGES.rst
mcoughlin: fixed

lpsinger (approved):

PR #573, opened 2026-09-05
Fix ground-based observer indexing in variable exposure time mode

mcoughlin:
This PR fixes ground-based observer indexing in variable exposure time mode.

### DIFF
--- a/src/m4opt/_cli/schedule.py
+++ b/src/m4opt/_cli/schedule.py
@@ -363,9 +363,7 @@ def schedule(
                 appmag_quantiles[np.isposinf(appmag_mu)] = np.inf
 
                 with observing(
-                    observer_location=observer_locations
-                    if isinstance(mission.observer_location, EarthFixedObserverLocation)
-                    else observer_locations[0],
+                    observer_location=observer_locations[0],
                     target_coord=hpx.healpix_to_skycoord(good)[:, np.newaxis],
                     obstime=obstimes[0],
                 ):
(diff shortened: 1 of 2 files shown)


src/m4opt/_cli/schedule.py
```diff
@@ -363,9 +363,7 @@ def schedule(
                 appmag_quantiles[np.isposinf(appmag_mu)] = np.inf
 
                 with observing(
-                    observer_location=observer_locations
-                    if isinstance(mission.observer_location, EarthFixedObserverLocation)
-                    else observer_locations[0],
+                    observer_location=observer_locations[0],
```
lpsinger, line 389:
```suggestion
                    observer_location=observer_locations[0]
                    if isinstance(mission.observer_location, EarthFixedObserverLocation)
                    else observer_locations,
```

src/m4opt/_cli/schedule.py
```diff
@@ -402,9 +400,7 @@ def schedule(
             else:
                 distmod = Distance(skymap_moc.meta["distmean"] * u.Mpc).distmod
                 with observing(
-                    observer_location=observer_locations
-                    if isinstance(mission.observer_location, EarthFixedObserverLocation)
-                    else observer_locations[0],
+                    observer_location=observer_locations[0],
```
lpsinger, line 426: Same as above.

lpsinger (changes requested):
That patch would break space-based observing.

I think that the `if` statement was just backwards. Do you agree?

mcoughlin:
@lpsinger I get:

ValueError: shape mismatch: objects cannot be broadcast to a single shape.
Mismatch is between arg 0 with shape (361,) and arg 1 with shape (2114,).

with your version.

lpsinger:
> @lpsinger I get:
> 
> ValueError: shape mismatch: objects cannot be broadcast to a single shape. Mismatch is between arg 0 with shape (361,) and arg 1 with shape (2114,).
> 
> with your version.

You're right. My mistake. Your patch is correct. Please fix the merge conflict.

lpsinger (approved):

PR #574, opened 2026-09-05
Schedule visits in per-bandpass blocks to limit filter changes

mcoughlin:
This PR schedules visits in per-bandpass blocks to limit filter changes (like is needed for ZTF).


lpsinger (commented):
What is the relationship of this work to @devas015' code from her internship?

mcoughlin:
@lpsinger This is meant to cover ZTF's preferences vs @devas015 who is working on Rubin. I suspect this would help her, although the Rubin problem is a bit different.

lpsinger:
> @lpsinger This is meant to cover ZTF's preferences vs @devas015 who is working on Rubin. I suspect this would help her, although the Rubin problem is a bit different.

Have you discussed this PR with her and compared with her code?

devas015:
To confirm I am understanding correctly- this code adds a constraint of a minimum "exchange time" between visits that have different filters, which should (but doesn't have to?) end up enforcing that each field is visited in one filter before being visited in another? My code from the internship explicitly enforces that latter part but does not take into account the time to change filters.

mcoughlin:
@devas015 can you send me your code or point to a branch / PR for me to look at?

The way it works is that we have a pair of variables per visit index, block_start[v] and block_end[v]. Every observed field's visit v is forced inside [block_start[v], block_end[v]] and then consecutive blocks are separated block_start[v+1] - block_end[v] >= exchange_time  (when the bandpass changes).

devas015:
[This branch](https://github.com/devas015/m4opt/tree/rubin-too-whitepaper) has how I implemented the visit blocks. `filter_change` refers to the enforcement of visiting every field k times before visiting any field k+1 times (disregard `preset_exp_del` as that refers to how constraints are added with a preset exposure time and delay sequence). 

Thank you for the explanation about the blocks- that makes more sense. Does adding the constraints of the blocks here raise any concerns about the problem becoming too complex for the solver?

mcoughlin:
@devas015 yes your way looks good. Can I incorporate your code into this PR to help get it to the finish line? I like yours as it has no new variables.

devas015:
@mcoughlin Absolutely! Let me know if you need me to rework anything.

mcoughlin:
thanks @devas015. I merged in your branch and tagged you as reviewer.

src/m4opt/_cli/schedule.py
devas015, line 256: Maybe add documentation for the new filter change mode?

devas015 (commented):
This looks good to me! Just one consideration - do we want a way to opt out of the visit blocking for telescopes with short filter changes (like UVEX)? Would enforcing the blocking on maps that don't necessarily need it negatively affect the schedule at all?

src/m4opt/_cli/schedule.py
mcoughlin: done

mcoughlin:
@devas015 I checked and if anything, the block schedule is faster (and of course makes no difference in the single filter case), so it should be ok as is.

lpsinger (commented):
Wow, this is very elegantly done with minimal impact to the existing code base. Taking a look now.

CHANGES.rst
```diff
@@ -78,6 +78,15 @@ Changes
 
 - Add ``intersect1d``, an accelerated version of ``numpy.intersect1d``.
 
+- Allow ``--bandpass`` to be repeated so that successive visits cycle through
+  several bandpasses. Every field is visited for the kth time before any field
+  is visited for the k+1th, so a schedule exchanges the filter once per block
+  boundary however many fields are observed. The ordering also makes the
+  no-overlap constraint redundant across visits, leaving an absolute value
+  only within one.
```
lpsinger: What does "leaving an absolute value only within one" mean?

CHANGES.rst
lpsinger, line 1: Move to current changelog entry.

docs/design/milp.rst
```diff
@@ -136,10 +136,43 @@ Maximize the sum of the probability of all of the pixels that are contained with
...
+The constraints are slightly different:
+
+**Depth.** Same as above.
+
+**Cadence.** Same as above.
+
+**No overlap.** This is similar to Equation :eq:`fixed-exptime-constraint-no-overlap`, but is broken down into two cases with slightly different left-hand sides.
+
+.. math::
+    :label: filter-change-constraint-no-overlap
```
lpsinger, line 159:
```suggestion
    :label: filter-change-constraint-no-overlap
    :no-wrap:
```

See https://github.com/sphinx-doc/sphinx/issues/14498.

docs/design/milp.rst
```diff
@@ -136,10 +136,43 @@ Maximize the sum of the probability of all of the pixels that are contained with
 
     \sum_{i \in I} \rho_i p_i
 
-Problem 2: Variable exposure time
+Problem 2: Filter changes with fixed exposure time
+--------------------------------------------------
+
+In this variation, we enforce that every field must be visited :math:`k` times before any field is visited :math:`k + 1` times to allow for length filter changes to occur between visits. The exposure time is fixed for all fields and visits.
```
lpsinger:
```suggestion
In this variation, we enforce that every field must be visited for the :math:`k`th before any field is visited for the :math:`(k + 1)` time to allow for length filter changes to occur between visits. The exposure time is fixed for all fields and visits.
```

src/m4opt/_cli/tests/test_filter_blocks.py
lpsinger, line 1: Could you test with ZTF? UVEX always takes images in both bands and doesn't have a filter changer.

src/m4opt/_cli/tests/test_filter_blocks.py
```diff
@@ -0,0 +1,89 @@
...
+
+def test_one_filter_change_per_block_boundary(schedule):
+    """The schedule exchanges the filter once per boundary, not once per field."""
+    if len(schedule) == 0:
+        pytest.skip("no fields were observable")
+    schedule.sort("start_time")
+    bands = list(schedule["bandpass"])
+    changes = sum(a != b for a, b in pairwise(bands))
+    assert changes == 2
+    assert len(schedule) > changes
```
lpsinger: Combine into a single test so you're not running the scheduler twice with exactly the same arguments.

src/m4opt/_cli/tests/test_filter_blocks.py
```diff
@@ -0,0 +1,89 @@
...
+        "--max-fields=6",
+        "--deadline=8hour",
+        "--timelimit=60s",
+        "--exptime-min=300s",
+        "--no-appmag-dist",
+    )
+    assert result.exit_code == 0
+    table = QTable.read(out)
+    observations = table[table["action"] == "observe"]
+    assert set(np.asarray(observations["bandpass"])) <= {"NUV"}
```
lpsinger:
```suggestion
    assert set(np.asarray(observations["bandpass"])) == {"NUV"}
```

src/m4opt/_cli/schedule.py
```diff
@@ -250,9 +273,24 @@ def schedule(
...
+    boundary however many fields are observed.
     """
     adaptive_exptime = absmag_mean is not None
 
+    # Successive visits cycle through the requested bandpasses, so that
+    # --bandpass g --bandpass r over three visits gives g, r, g.
+    visit_bandpasses = [
+        bandpass[i % len(bandpass)] if bandpass else None for i in range(visits)
+    ]
+    unique_bandpasses = list(dict.fromkeys(visit_bandpasses))
```
lpsinger: I understand that this idiom gives you the unique elements in order of their first appearance. However it is a little opaque. Please factor it out as its own function and give it a descriptive name, such as `unique_preserving_order`.

Is it important to preserve the order?

src/m4opt/_cli/schedule.py
```diff
@@ -250,9 +273,24 @@ def schedule(
        Gaussian distribution of absolute magnitudes. This mode is selected if
        you specify the --absmag-mean option (and, optionally, the
        --absmag-stdev option).
+
+    \b
+    Repeating the --bandpass option makes successive visits cycle through the
+    bandpasses; for example, --bandpass g --bandpass r observes each field in g
+    and then in r. Visits are grouped into contiguous blocks of a single
+    bandpass, so that every field is observed for the kth time before any field
+    is observed for the k+1th, and the filter is exchanged once per block
```
lpsinger:
```suggestion
    is observed for the (k+1)th, and the filter is exchanged once per block
```

src/m4opt/_cli/schedule.py
```diff
@@ -250,9 +273,24 @@ def schedule(
        Gaussian distribution of absolute magnitudes. This mode is selected if
        you specify the --absmag-mean option (and, optionally, the
        --absmag-stdev option).
+
+    \b
+    Repeating the --bandpass option makes successive visits cycle through the
+    bandpasses; for example, --bandpass g --bandpass r observes each field in g
+    and then in r. Visits are grouped into contiguous blocks of a single
```
lpsinger:
```suggestion
    bandpasses; for example, --bandpass g --bandpass r observes every field in g
    and then every field in r. Visits are grouped into contiguous blocks of a single
```

src/m4opt/_cli/schedule.py
```diff
@@ -82,6 +83,21 @@ def prepare_piecewise_breakpoints(breakpoints):
...
+
+    A field has a single exposure time shared by all of its visits, so the
+    least sensitive bandpass governs.
+    """
+    return np.maximum.reduce(
+        [
+            mission.detector.get_exptime(snr, spectrum, bandpass).to_value(u.s)
+            for bandpass in bandpasses
+        ]
+    )
```
lpsinger: Is this patch intended to support both fixed and adaptive exposure time while using a filter exchanger?

If so, why would we lock all of the filters to use the same exposure time?

lpsinger (changes requested):

src/m4opt/_cli/schedule.py
mcoughlin: Yeah I got the fixed part of this. Adaptive is a bigger change and will take more work.

src/m4opt/_cli/schedule.py
mcoughlin: The order is nice for keeping the runs same each time, but otherwise no

CHANGES.rst
mcoughlin: yes.

CHANGES.rst
mcoughlin: tried to clarify this.

src/m4opt/_cli/tests/test_filter_blocks.py
mcoughlin: done

src/m4opt/_cli/schedule.py
lpsinger: It's fine to keep them in order, but please refactor into a function.

mcoughlin: _unique_preserving_order is called from here. Maybe you meant a bigger refactor?

src/m4opt/_cli/schedule.py
lpsinger: In that case, I'd rather just raise a NotImplementedError if the user supplies both the adaptive exposure time options and the filter exchanger option.

lpsinger (changes requested):

src/m4opt/_cli/schedule.py
lpsinger: Oops!

src/m4opt/_cli/animate.py
```diff
@@ -220,6 +226,12 @@ def animate(
                 raise NotImplementedError(
                     "This mission does not define a detector model"
                 )
+            if len(bandpass) > 1:
+                raise NotImplementedError(
+                    "An exposure time map is not supported for a schedule that "
+                    "uses more than one bandpass, because each field has a "
+                    "single exposure time that every one of its visits shares."
+                )
```
lpsinger: This is too much information for the exception.

```suggestion
                    "uses more than one bandpass."
                )
```

And then let's create an issue to add adaptive exposure time support for filter changes.

src/m4opt/_cli/animate.py
```diff
@@ -92,6 +93,11 @@ def animate(
         visits = table.meta["args"]["visits"]
         absmag_mean = table.meta["args"]["absmag_mean"]
         bandpass = table.meta["args"]["bandpass"]
+        # Schedules written before the bandpass became one entry per visit
+        # record a single name.
+        if isinstance(bandpass, str) or bandpass is None:
+            bandpass = [bandpass]
+        bandpass = _unique_preserving_order(bandpass)
```
lpsinger: In this script, you only care about the zeroth bandpass, which is the same whether you take it from the original list or the deduplicated list. So there is no need for _unique_preserving_order here.

src/m4opt/_cli/animate.py
```diff
@@ -33,6 +33,7 @@
 from ..synphot.extinction import DustExtinction
 from ..utils.console import progress, status
 from .core import app
+from .schedule import _unique_preserving_order
```
lpsinger: If you need this method across multiple CLI entry points, then a better place for it would be m4opt.utils. However, I don't think you do.

src/m4opt/_cli/schedule.py
```diff
@@ -130,11 +175,13 @@ def schedule(
...
-        u.Quantity,
+        list[u.Quantity] | None,
         typer.Option(
-            help="Minimum exposure time for each observation",
+            help="Minimum exposure time for each observation. Repeat the "
+            "option to give each bandpass its own exposure time, in the same "
+            "order as --bandpass; a single value applies to every bandpass",
         ),
-    ] = 900 * u.s,
+    ] = None,
```
lpsinger: This should still be a required option, with no default. Changing the type from `u.Quantity` to `list[u.Quanity]` should be sufficient to accept multiple occurrences of the option.

```suggestion
        list[u.Quantity],
typer.Option(
            help="Minimum exposure time for each observation. Repeat the "
            "option to give each bandpass its own exposure time, in the same "
            "order as --bandpass; a single value applies to every bandpass",
),
    ],
```

src/m4opt/_cli/schedule.py
```diff
@@ -82,6 +83,50 @@ def prepare_piecewise_breakpoints(breakpoints):
...
 
 
+def _exptime_min_per_visit(exptime_min, bandpass, visit_bandpasses, adaptive_exptime):
+    """
+    Minimum exposure time for each visit, one entry per visit.
+
+    ``--exptime-min`` may be given once, applying to every bandpass, or once per
+    bandpass in the order the bandpasses were given.
+    """
+    values = list(exptime_min) if exptime_min else [900 * u.s]
```
lpsinger: There is no good default value for --exptime-min.

src/m4opt/_cli/schedule.py
```diff
@@ -82,6 +83,50 @@ def prepare_piecewise_breakpoints(breakpoints):
...
+    # A None default costs the option its unit checking, so check it here.
+    try:
+        values = [value.to(u.s) for value in values]
+    except u.UnitConversionError as e:
+        raise UsageError(f"--exptime-min must be a time: {e}") from None
```
lpsinger: Get rid of the None default (see below).

src/m4opt/_cli/schedule.py
```diff
@@ -82,6 +83,50 @@ def prepare_piecewise_breakpoints(breakpoints):
...
+
+    if len(values) == 1:
+        return [values[0]] * len(visit_bandpasses)
+
+    bandpasses = bandpass or []
+    if len(values) != len(bandpasses):
+        raise UsageError(
+            f"Got {len(values)} values for --exptime-min and {len(bandpasses)} "
+            "for --bandpass. Give one exposure time in total, or one for every "
```
lpsinger:
```suggestion
            "for --bandpass. Give exactly one exposure time, or one for every "
```

src/m4opt/_cli/schedule.py
```diff
@@ -82,6 +83,50 @@ def prepare_piecewise_breakpoints(breakpoints):
...
+            "bandpass."
+        )
+    if adaptive_exptime:
+        raise UsageError(
+            "A separate --exptime-min for each bandpass is supported only with "
+            "a fixed exposure time. Omit --absmag-mean, or give a single "
+            "--exptime-min."
+        )
+    by_bandpass = dict(zip(bandpasses, values))
+    return [by_bandpass[band] for band in visit_bandpasses]
```
lpsinger: Since we require there to be exactly the same number of `--bandpass` arguments as `--exptime-min` arguments, then the user probably expects that each visit will be in the _i_th bandpass with the _i_th exposure time.

So this shouldn't pass through a dictionary lookup.

src/m4opt/_cli/schedule.py
```diff
@@ -250,9 +304,36 @@ def schedule(
...
+    # --bandpass g --bandpass r over three visits gives g, r, g.
+    visit_bandpasses = [
+        bandpass[i % len(bandpass)] if bandpass else None for i in range(visits)
+    ]
+    unique_bandpasses = _unique_preserving_order(visit_bandpasses)
+    visit_exptime_min = _exptime_min_per_visit(
+        exptime_min, bandpass, visit_bandpasses, adaptive_exptime
+    )
+    # The shortest of them bounds anything that needs a single number.
+    shortest_exptime_min = min(visit_exptime_min)
```
lpsinger: I think that the only places that need a single number are those places that can't be reached since we aren't implementing adaptive exposure time with the filter changer.

lpsinger (changes requested):

src/m4opt/_cli/schedule.py
mcoughlin: done

src/m4opt/_cli/animate.py
mcoughlin: yup.

src/m4opt/_cli/animate.py
mcoughlin: yup.

src/m4opt/_cli/schedule.py
mcoughlin: added this.

src/m4opt/_cli/schedule.py
mcoughlin: added.

src/m4opt/_cli/schedule.py
mcoughlin: done

src/m4opt/_cli/schedule.py
mcoughlin: yup

src/m4opt/_cli/schedule.py
mcoughlin: yes

src/m4opt/_cli/schedule.py
```diff
@@ -250,9 +294,32 @@ def schedule(
...
     """
     adaptive_exptime = absmag_mean is not None
 
+    # Successive visits cycle through the requested bandpasses, so that
+    # --bandpass g --bandpass r over three visits gives g, r, g.
+    visit_bandpasses = [
+        bandpass[i % len(bandpass)] if bandpass else None for i in range(visits)
+    ]
+    unique_bandpasses = _unique_preserving_order(visit_bandpasses)
+    visit_exptime_min = _exptime_min_per_visit(exptime_min, bandpass, visits)
```
lpsinger: Here again, I don't think you need to track the unique bandpasses because we aren't supporting the adaptive exposure time with multiple bandpasses and because the number of `--bandpass` arguments must be exactly equal to the number of `--exptime-min` arguments. You're going to cycle through the exposure times and bandpasses in lock step.

lpsinger (changes requested):

src/m4opt/_cli/schedule.py
mcoughlin: yeah this can go completely, removing.

src/m4opt/_cli/animate.py
```diff
@@ -92,8 +92,13 @@ def animate(
         visits = table.meta["args"]["visits"]
         absmag_mean = table.meta["args"]["absmag_mean"]
         bandpass = table.meta["args"]["bandpass"]
+        # Schedules written before the bandpass became one entry per visit
```
lpsinger:
```suggestion
        # Schedules written before https://github.com/m4opt/m4opt/pull/574
```

src/m4opt/_cli/schedule.py
```diff
@@ -82,6 +83,30 @@ def prepare_piecewise_breakpoints(breakpoints):
...
+
+    ``--exptime-min`` is given either once, applying to every visit, or once for
+    every ``--bandpass``, pairing with them in the order both were given.
+    """
+    # The option is required and so has no default, which costs it the unit
+    # checking that a default would have supplied.
+    try:
+        values = [value.to(u.s) for value in exptime_min]
+    except u.UnitConversionError as e:
+        raise UsageError(f"--exptime-min must be a time: {e}") from None
```
lpsinger: No longer necessary because the default value is no longer None.

src/m4opt/_cli/schedule.py
```diff
@@ -250,9 +284,28 @@ def schedule(
...
+    boundary however many fields are observed.
     """
     adaptive_exptime = absmag_mean is not None
 
+    # Successive visits cycle through the requested bandpasses, so that
+    # --bandpass g --bandpass r over three visits gives g, r, g.
+    visit_bandpasses = [
+        bandpass[i % len(bandpass)] if bandpass else None for i in range(visits)
+    ]
+    visit_exptime_min = _exptime_min_per_visit(exptime_min, bandpass, visits)
```
lpsinger:
```suggestion
    visit_exptime_min_s = u.Quantity([
        exptime_min[i % len(exptime_min)] for i in range(visits)
    ]).to_value(u.s)
```

src/m4opt/_cli/schedule.py
```diff
@@ -293,7 +346,10 @@ def schedule(
         target_coords = target_coords.unmasked[keep]
         # FIXME: https://github.com/astropy/astropy/issues/17030
         target_coords = SkyCoord(target_coords.ra, target_coords.dec)
-        exptime_min_s = exptime_min.to_value(u.s)
+        exptime_min_s = min(visit_exptime_min).to_value(u.s)
+        visit_exptime_min_s = np.array(
+            [value.to_value(u.s) for value in visit_exptime_min]
+        )
```
lpsinger:
```suggestion
```

src/m4opt/_cli/schedule.py
```diff
@@ -82,6 +83,30 @@ def prepare_piecewise_breakpoints(breakpoints):
     return [tuple(col.item() for col in row) for row in breakpoints]
 
 
+def _exptime_min_per_visit(exptime_min, bandpass, visits):
```
lpsinger: I think you can probably delete this function due to the simplifications of the CLI below.

src/m4opt/_cli/schedule.py
```diff
@@ -402,8 +458,7 @@ def schedule(
                     target_coord=hpx.healpix_to_skycoord(good)[:, np.newaxis],
                     obstime=obstimes[0],
                 ):
-                    exptime_pixel_s = mission.detector.get_exptime(
-                        snr,
+                    spectrum = (
```
lpsinger: It wasn't necessary to pull out `spectrum` here, was it? Please undo to minimize the diff.

lpsinger (changes requested):

src/m4opt/_cli/schedule.py
mcoughlin: sorry about that.

src/m4opt/_cli/schedule.py
mcoughlin: yes

src/m4opt/_cli/schedule.py
mcoughlin: yes.

src/m4opt/_cli/core.py
lpsinger, line 1: Would you please pull out the changes to the core CLI functions and tests as a separate PR?

lpsinger (changes requested):

src/m4opt/_cli/core.py
lpsinger: I feel bad asking you to split the PR, because the changes are in response to my comments. Would you like me to open that separate PR for you?

mcoughlin: I did it

lpsinger: Thank you! You didn't have to remove the code here, though. I want to get #608 finished and merged, then rebase this on top of it, then finish reviewing this and merge.

lpsinger: Now that #608 is merged, please rebase.

mcoughlin: @lpsinger done

src/m4opt/_cli/schedule.py
```diff
@@ -95,6 +96,15 @@ def schedule(
             help="Output filename for generated schedule", metavar="SCHEDULE.ecsv"
         ),
     ],
+    exptime_min: Annotated[
+        list[u.Quantity],
+        u.physical.time,
```
lpsinger:
```suggestion
        list[u.Quantity[u.physical.time]],
```

src/m4opt/_cli/schedule.py
```diff
@@ -534,31 +566,85 @@ def schedule(
             if visits > 1:
                 with status("adding cadence constraints"):
                     if adaptive_exptime:
-                        rhs = cadence_s * field_vars + exptime_field_vars
+                        rhs = (cadence_s * field_vars + exptime_field_vars)[
+                            :, np.newaxis
+                        ]
                     else:
-                        rhs = (exptime_min_s + cadence_s) * field_vars
+                        half_sum = 0.5 * (
```
lpsinger: Where does this new factor of 1/2 come from?

CHANGES.rst
```diff
@@ -5,6 +5,25 @@ Changes
 2.16.0 (unreleased)
 ===================
 
+- Allow ``--bandpass`` to be repeated so that successive visits cycle through
+  several bandpasses. Every field is visited for the kth time before any field
+  is visited for the k+1th, so a schedule exchanges the filter once per block
+  boundary however many fields are observed. Ordering the visits also fixes the
+  sign of the time difference between two observations that are in different
+  visits, so the no-overlap constraint needs an absolute value only between two
+  observations within the same visit.
```
lpsinger: The change log shouldn't mention implementation details.

```suggestion
  boundary however many fields are observed.
```

CHANGES.rst
```diff
@@ -5,6 +5,25 @@ Changes
...
+
+- Add ``Mission.filter_exchange_time`` and set it to 110 s for ZTF.
+
+- Allow ``--exptime-min`` to be repeated so that each bandpass has its own
+  exposure time, given in the same order as ``--bandpass``; a single value
+  applies to every bandpass as before. The scheduler spaces two observations by
+  the slew time plus half of each of their exposure times, so a short exposure
+  in one filter no longer has to be padded out to the length that the least
+  sensitive filter needs. This applies to a fixed exposure time; with
```
lpsinger: Irrelevant. We never supported different exposure times for the same field before.

```suggestion
  applies to every bandpass as before. This applies to a fixed exposure time; with
```

src/m4opt/_cli/schedule.py
```diff
@@ -534,31 +566,85 @@ def schedule(
...
+                            visit_exptime_min_s[:-1] + visit_exptime_min_s[1:]
+                        )
+                        rhs = np.multiply.outer(field_vars, cadence_s + half_sum)
                     model.add_constraints_(
                         (time_field_visit_vars[:, 1:] - time_field_visit_vars[:, :-1])
-                        >= rhs[:, np.newaxis]
+                        >= rhs
                     )
 
             with status("adding slew constraints"):
```
lpsinger, line 588: This will take me a little longer to check because it is a complete rewrite.

lpsinger (changes requested):

src/m4opt/_cli/schedule.py
mcoughlin: It isn't new, the slew constraint already used the same thing. Times are midpoints, so visit k occupies [t_k − E_k/2, t_k + E_k/2]. The edge-to-edge gap between consecutive visits is therefore (t_{k+1} − E_{k+1}/2) − (t_k + E_k/2)  =  Δt − (E_k + E_{k+1})/2 and requiring that to be at least the cadence gives Δt ≥ cadence + (E_k + E_{k+1})/2.

lpsinger: In the original code, this `if: else:` block is _only_ calculating the right hand side for the `model.add_constraints_` call that immediately follows it, but this `half_sum` definition is unrelated to this constraint; it is used in the `"adding slew constraints"` section. Please move it near to where it is actually used.

lpsinger (changes requested):

mcoughlin:
@lpsinger the same expression appeared again inline in the slew section so now it is above both as mean_consecutive_exptime_s

CHANGES.rst
```diff
@@ -5,6 +5,19 @@ Changes
...
+  is visited for the k+1th, so a schedule exchanges the filter once per block
+  boundary however many fields are observed.
+
+- Add ``Mission.filter_exchange_time`` and set it to 110 s for ZTF.
+
+- Allow ``--exptime-min`` to be repeated so that each bandpass has its own
+  exposure time, given in the same order as ``--bandpass``; a single value
+  applies to every bandpass as before. This applies to a fixed exposure time; with
+  ``--absmag-mean`` the scheduler now refuses more than one bandpass rather
+  than quietly making every filter share one exposure time.
```
lpsinger:
```suggestion
  applies to every bandpass as before. This applies to a fixed exposure time mode
  and is not yet supported with adaptive exposure time mode.
```

lpsinger (approved):

lpsinger (changes requested):
Please address the final changelog comment.

lpsinger (approved):

lpsinger:
Do you need a release at this point, or will you be doing some more testing first?

mcoughlin:
@lpsinger please do a release. :)

lpsinger:
> @lpsinger please do a release. :)

Done.

PR #576, opened 2026-09-05
[WIP] Add SCIP as an open-source MILP solver backend

mcoughlin:
Extra confusing as it builds on top of https://github.com/m4opt/m4opt/pull/459.


lpsinger (commented):
See https://github.com/m4opt/m4opt/pull/459#pullrequestreview-5127416950.

PR #578, opened 2026-09-05
event-time option

mcoughlin:
This PR adds --event-time to set the time.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,8 +18,8 @@ Changes
   schedules were empty as a result.
 
 - Add ``--event-time``, the time that ``--delay`` and ``--deadline`` are
-  measured from. It defaults to the ``gps_time`` in the sky map header as
-  before, so a sky map generated without one can now be scheduled, and a sky
+  measured from. It defaults to the ``DATE-OBS`` field of the sky map as
+  before, so a sky map written without one can now be scheduled, and a sky
   map missing it reports what to do rather than a ``KeyError``. Schedules
   record the time they were measured from, and ``m4opt animate`` reads it
   from there rather than from the sky map.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,11 @@ Changes
   table reader so that all 1778 fields fell within 10 degrees of R.A. 0. ZTF
   schedules were empty as a result.
 
+- Add ``--event-time``, the time that ``--delay`` and ``--deadline`` are
+  measured from. It defaults to the ``gps_time`` in the sky map header as
+  before, so a sky map generated without one can now be scheduled, and a sky
+  map missing it reports what to do rather than a ``KeyError``.
+
 - Use ULTRASAT's tabulated throughput curve rather than a Gaussian
   approximation, which had a red leak some four orders of magnitude too
   large and nearly doubled the predicted zodiacal background.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,7 +20,9 @@ Changes
 - Add ``--event-time``, the time that ``--delay`` and ``--deadline`` are
   measured from. It defaults to the ``gps_time`` in the sky map header as
   before, so a sky map generated without one can now be scheduled, and a sky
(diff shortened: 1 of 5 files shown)


lpsinger (changes requested):
What kind of a madman would produce a localization FITS file without an obstime?

The `m4opt animate` script might also be affected.

src/m4opt/_cli/schedule.py
```diff
@@ -100,6 +100,13 @@ def schedule(
             help="Name of sky grid to use, if the mission supports multiple sky grids.",
         ),
     ] = None,
+    event_time: Annotated[
+        Time | None,
+        typer.Option(
+            help="Time of the event, which --delay and --deadline are measured "
+            "from. Defaults to the gps_time in the sky map header.",
```
lpsinger:
```suggestion
            "from. Defaults to the DATE-OBS field in the sky map header.",
```

GPS times are not stored in FITS headers. The date is stored in the `DATE-OBS` field. The `ligo.skymap.io.read_sky_map` method parses that field and returns the GPS time.

See https://git.ligo.org/lscsoft/ligo.skymap/-/blob/main/ligo/skymap/io/fits.py?ref_type=heads#L221-227).

src/m4opt/_cli/animate.py
```diff
@@ -93,14 +94,26 @@ def animate(
         bandpass = table.meta["args"]["bandpass"]
         snr = table.meta["args"]["snr"]
         exptime_min = table.meta["args"]["exptime_min"]
+        # The schedule records the time it was measured from.
+        recorded_event_time = table.meta["args"].get("event_time")
```
lpsinger:
```suggestion
        event_time = table.meta["args"].get("event_time")
```

src/m4opt/_cli/animate.py
```diff
@@ -93,14 +94,26 @@ def animate(
...
 
     with status("loading sky map"):
         hpx = HEALPix(nside, frame=ICRS(), order="nested")
         skymap_moc = read_sky_map(skymap, moc=True)
         probs = rasterize(skymap_moc["UNIQ", "PROBDENSITY"], hpx.level)["PROB"]
-        event_time = Time(
-            Time(skymap_moc.meta["gps_time"], format="gps").utc, format="iso"
-        )
+        if recorded_event_time is not None:
```
lpsinger:
```suggestion
        if recorded_event_time is None:
```

and switch the `if` and `else` statements. I prefer to avoid the double negative `if not ...: else ...:`.

lpsinger (changes requested):

src/m4opt/_cli/schedule.py
```diff
@@ -246,9 +253,16 @@ def schedule(
...
-        event_time = Time(
-            Time(skymap_moc.meta["gps_time"], format="gps").utc, format="iso"
-        )
+        if event_time is None:
+            # The sky map carries the trigger time unless one was given.
+            try:
+                gps_time = skymap_moc.meta["gps_time"]
+            except KeyError:
+                raise UsageError(
+                    f'The sky map "{skymap.name}" has no gps_time in its header. '
```
lpsinger: Update this error message.

CHANGES.rst
```diff
@@ -17,6 +17,13 @@ Changes
   table reader so that all 1778 fields fell within 10 degrees of R.A. 0. ZTF
   schedules were empty as a result.
 
+- Add ``--event-time``, the time that ``--delay`` and ``--deadline`` are
+  measured from. It defaults to the ``gps_time`` in the sky map header as
```
lpsinger: DATE-OBS

lpsinger (changes requested):

src/m4opt/_cli/schedule.py
mcoughlin: Fixed.

CHANGES.rst
mcoughlin: Fixed

lpsinger (approved):

lpsinger (approved):

PR #579, opened 2026-09-05
max-fields

mcoughlin:
This PR adds --max-fields to control how many fields are considered

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,10 @@ Changes
 2.12.1 (unreleased)
 ===================
 
+- Add ``--max-fields``, the number of most probable fields to consider. It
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,10 @@ Changes
 2.12.1 (unreleased)
 ===================
 
+- Add ``--max-fields``, the number of most probable fields to consider. It
+  was fixed at 50, which silently truncated the problem for instruments whose
+  sky grid covers a localization with more fields than that.
```
lpsinger: I don't understand this explanation. It is normal that the "sky grid covers a localization with more fields than" 50.

lpsinger (changes requested):

lpsinger (approved):

PR #580, opened 2026-09-05
Differentiate whether a schedule was solved or not vs unobservable

mcoughlin:
This PR differentiates whether a schedule was solved or not vs unobservable.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,10 @@ Changes
 - Fix the ULTRASAT readout noise, which was set to the noise budget's
   variance (6 e-/pix) rather than its RMS.
 
+- Record ``has_observable_fields`` and ``has_solution`` in the schedule
+  metadata. An empty schedule previously gave no way to tell a sky that was
+  never observable from a solve that ran out of time.
+
 2.12.0 (2026-08-28)
 ===================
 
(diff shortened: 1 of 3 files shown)


lpsinger (commented):
Here is (as far as I can tell) an exhaustive list of the reasons that the schedule may be empty:

- The best bound dropped below the `--cutoff` value. The scheduler has proved that no possible schedule would meet the cutoff. This should be apparent in the `best_bound` metadata value.
- The solver reached the time limit while `best_bound` was still greater than `--cutoff`, but without having found any feasible solutions. This should be apparent from the value of the `solution_status` metadata value; it should tell you that the time limit was exceeded.

Did I miss anything above? Are there any empty schedules that do not fit into either of the above categories?

mcoughlin:
@lpsinger I think there are a couple more (these may be versions of subsets of yours, but more specific)
(iii) A solve that succeeds and returns nothing, e.g. the empty selection is the only feasible one.
(iv) a sky that is never observable.

lpsinger:
> @lpsinger I think there are a couple more (these may be versions of subsets of yours, but more specific)
> (iii) A solve that succeeds and returns nothing, e.g. the empty selection is the only feasible one.
> (iv) a sky that is never observable.

I think that they are subsets.

Do you have some example inputs and outputs for which the metadata appears ambiguous? An example would help me to understand how to make this more usable.

PR #588, opened 2026-09-10
Apply short-circuit evaluation to logical constraints

lpsinger:


mcoughlin (changes requested):
Can you add a test where np.take works through ShapedLikeNDArray (e.g. one where there is one composite constraint over a two dimensional grid)?

lpsinger:
> Can you add a test where np.take works through ShapedLikeNDArray (e.g. one where there is one composite constraint over a two dimensional grid)?

That is already exercised in test_end_to_end.py.

Also thanks to the magic of hypothesis' mutually_broadcastable_shapes, the unit tests already cover many examples of N-D grids.

mcoughlin (approved):
LGTM!

PR #593, opened 2026-09-11
Add zodiacal background constraint

lpsinger:


src/m4opt/constraints/_zodiacal.py
```diff
@@ -0,0 +1,71 @@
...
+            ax.set_ylabel(r"Ecliptic latitude, $\beta$")
+            ax.grid()
+    """
+
+    max_surface_brightness: float
+    r"""Maximum visual surface brightness, :math:`m_V \, \mathrm{arcsec}^{-2}`."""
+
+    @override
+    def __call__(self, *args):
+        return mag_at(*args) <= self.max_surface_brightness
```
mcoughlin: I think this is backwards (since magnitudes are weird) :)

mcoughlin (changes requested):
I think your inequality is inverted

src/m4opt/constraints/_zodiacal.py
lpsinger: Dammit.

Once again, when I rule the world, I will decree that astronomers shall use decibels instead of magnitudes. Yes, it's another stupid logarithmic scale, but at least it goes in the right direction, and is used in other fields.

mcoughlin: lolz

lpsinger: Should be fixed now. I mean the code, not the astronomy profession's slavish devotion to ancient units of measure. The later is still WIP.

mcoughlin (approved):
LGTM

mcoughlin (approved):
LGTM

PR #598, opened 2026-09-13
Add NominalRollConstraint

lpsinger:


src/m4opt/constraints/_positional.py
```diff
@@ -47,45 +73,40 @@ def _get_angle(
...
+class ICRSConstraint(FrameAngleConstraint):
     """Constrain an angle in the :class:`~astropy.coordinates.ICRS` frame."""
 
     @override
     def _frame(self, *_):
         return ICRS()
 
 
-class LongitudeConstraint(AngleConstraint):
+class LongitudeConstraint(FrameAngleConstraint, WrappedAngleConstraint):
```
mcoughlin, line 110:
```suggestion
@dataclass
class LongitudeConstraint(FrameAngleConstraint, WrappedAngleConstraint):
```

mcoughlin: This is needed for min to be wrapped

mcoughlin (changes requested):
Missing @dataclass on longitude constraint, otherwise should be good.

lpsinger:
> Missing @DataClass on longitude constraint, otherwise should be good.

Why do you say so? It doesn't add any fields.

src/m4opt/constraints/_positional.py
lpsinger: You're right! I added some regression tests for longitude constraints with max < min.

mcoughlin (approved):
LGTM

PR #602, opened 2026-09-14
Add roll angle field graphic

lpsinger:


lpsinger:
There's one thing I am still finding confusing. How does the roll angle "know" to go undefined at the ecliptic poles, as opposed to anywhere else within the plane at right angles to the Sun? There is only one defined direction in this problem: the direction to the Sun. Why isn't the roll angle field rotationally symmetric around that?

m4opt-dynamics-nominal_roll-1

mcoughlin (approved):
LGTM. The attitude is rotationally symmetric about the Sun, but you are plotting the position angle measured from celestial north, so the celestial pole is a second preferred direction in those units.

lpsinger:
> LGTM. The attitude is rotationally symmetric about the Sun, but you are plotting the position angle measured from celestial north, so the celestial pole is a second preferred direction in those units.

Do you think the following is correct? The Sun and anti-Sun directions are true singularities where the nominal roll angle and spacecraft orientation are undefined. The ecliptic poles are merely coordinate singularities where the nominal roll angle is undefined but the actual spacecraft orientation is defined.

mcoughlin:
@lpsinger Yes but the celestial poles, not the ecliptic poles, I think.

lpsinger:
> @lpsinger Yes but the celestial poles, not the ecliptic poles, I think.

You're absolutely right! I updated the text accordingly.

mcoughlin (approved):
LGTM

PR #604, opened 2026-09-16
Neglect nutation in zodiacal light calculation

lpsinger:
Calculate the zodiacal light background in the mean ecliptic frame rather than the true eclpitic frame. This results in much faster ETC calculations at the expense of positional accuracy degrading to a fraction of an arcminute due to neglecting nutation.


mcoughlin (approved):
LGTM

PR #605, opened 2026-09-17
Pregenerate Rubin FOV region file

lpsinger:
Parsing the YAML file is actually quite time consuming. Pregenerate the DS9 region file.

This also introduces a pytest fixture called generated_file which we can reuse to pregenerate and test other data products.


mcoughlin (approved):
LGTM

